PR #47, opened 2026-05-01
feat: add in-memory caching for GitHub API responses

### LINKED ISSUE #46: Add caching for GitHub API responses
## Description
Currently, GitHub fetches in the backend are executed as fresh requests, even when the same username is requested repeatedly within a short time window.

This causes:
- Slower response times  
- Increased GitHub API usage  
- Higher chance of rate-limit hits  

## Problem
There is no short-term server-side cache for GitHub fetch results. Repeated requests for the same username trigger redundant outbound GitHub API calls.

## Proposed Solution
Add a short-term in-memory cache for GitHub fetch results.

- Cache by key (username + fetch type/endpoint context)  
- Use TTL of 5–10 minutes  
- Support `refresh=true` to bypass cache and force fresh fetch  

## Expected Behavior
- Repeated same-username requests within TTL return cached data  
- Redundant GitHub API calls are reduced  
- `refresh=true` always fetches fresh data  

## Acceptance Criteria
- [ ] Same username within TTL returns cached result  
- [ ] GitHub API call count decreases for repeated requests  
- [ ] `refresh=true` bypasses cache and fetches fresh data  
- [ ] Cache expires correctly after TTL  
- [ ] No API response schema changes  
- [ ] Tests cover:
  - cache hit  
  - cache miss  
  - refresh bypass  
  - cache expiry  

## Scope
- Backend only (server/API layer)  
- No frontend/UI changes  
- No changes to existing response structure  

## Suggested Implementation
- Add cache utility (`lib/cache.ts`)  
- Integrate cache in `lib/github.ts` fetch functions  
- Use in-memory `Map` with expiry timestamps

paanipoorie:
## Summary
Adds short-term in-memory caching for GitHub API responses to reduce redundant API calls and improve performance.

## Changes
- Added cache utility (`lib/cache.ts`)
- Integrated caching into GitHub fetch logic
- TTL set to 5–10 minutes
- Added support for `refresh=true` to bypass cache

## Behavior
- Same username within TTL returns cached data
- Reduces repeated GitHub API calls
- `refresh=true` forces fresh fetch

## Test
- Ran `npm run format`, `npm run lint`, and `npm run test`
- Verified cache hit/miss behavior locally

Fixes #46

### DIFF
--- a/app/api/github/route.ts
+++ b/app/api/github/route.ts
@@ -4,18 +4,22 @@ import { getFullDashboardData } from '@/lib/github';
 export async function GET(request: Request) {
   const { searchParams } = new URL(request.url);
   const username = searchParams.get('username');
+  const refresh = searchParams.get('refresh') === 'true';
 
   if (!username) {
     return NextResponse.json({ error: 'Username is required' }, { status: 400 });
   }
 
   try {
-    const data = await getFullDashboardData(username);
+    const data = await getFullDashboardData(username, { bypassCache: refresh });
+    const cacheControl = refresh
+      ? 'no-cache, no-store, must-revalidate'
+      : 's-maxage=3600, stale-while-revalidate';
 
     return NextResponse.json(data, {
       status: 200,
       headers: {
-        'Cache-Control': 's-maxage=3600, stale-while-revalidate',
+        'Cache-Control': cacheControl,
       },
     });
   } catch (error: unknown) {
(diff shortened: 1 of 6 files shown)


paanipoorie:
Implemented caching with TTL and refresh bypass.

All checks are passing (format, lint, tests).  
Ready for review.

JhaSourav07:
@paanipoorie 

Its fine i am merging this PR.

PR #222, opened 2026-05-20
fix(calculate): handle single-day and empty calendars safely in streak calculations

### LINKED ISSUE #220: [BUG] calculateStreak throws TypeError and crashes on single-day calendars with 0 contributions
**Description:**

**Behavior Overview**

- A crash occurs in the streak calculation logic (lib/calculate.ts) when processing a contribution calendar containing only a single day if that day has zero contributions.
- Specifically, when there is only one day in the array, todayIndex is 0. The logic tries to resolve the previous day with const yesterday = days[todayIndex - 1]. Because todayIndex - 1 resolves to -1, yesterday becomes undefined. When the code subsequently attempts to evaluate if the streak is still alive using the grace period check (yesterday.contributionCount > 0), it throws a TypeError: Cannot read property 'contributionCount' of undefined and causes the API route to fail with a 500 Internal Server Error.

**Steps to Reproduce**

Invoke the API or calling utility with a contribution calendar that spans only a single day, where that day has 0 contributions. E.g. json:
```
{
  "totalContributions": 0,
  "weeks": [
    {
      "contributionDays": [
        {
          "contributionCount": 0,
          "date": "2024-01-01"
        }
      ]
    }
  ]
}
```

The server will throw: TypeError: Cannot read property 'contributionCount' of undefined

**Proposed Fix**
Add an explicit boundary check to safely resolve yesterday, and use optional chaining or a fallback value when checking yesterday.contributionCount

Animesh-86:
## Description
When calculating the user's current streak with the grace-period logic in lib/calculate.ts, the function resolves yesterday relative to the last index of the contribution days:

```
const todayIndex = days.length - 1;
const today = days[todayIndex];
const yesterday = days[todayIndex - 1];
```
If a user has a contribution calendar containing only 1 day (e.g. a brand-new user account, or dynamic range filters):

- todayIndex is 0.
- yesterday evaluates to days[-1], which is undefined.
- If today.contributionCount is 0, the logic attempts to check if the streak is alive by checking yesterday.contributionCount > 0.
- Since yesterday is undefined, accessing .contributionCount throws TypeError: Cannot read properties of undefined (reading 'contributionCount') and crashes the API server, returning a 500 Internal Server Error.

A similar crash occurs if days.length is 0 (empty list) because today is also undefined.

**Solution**

- Safety Boundary Guard: Introduced an early-return check if todayIndex < 0 to safely handle empty calendars by returning baseline stats.
- Defensive Resolution: Safely resolved yesterday with a ternary boundary check: todayIndex > 0 ? days[todayIndex - 1] : null.
- Optional Grace Check: Safely evaluated yesterday inside the isStreakAlive boolean expression:

`const isStreakAlive = today.contributionCount > 0 || (yesterday ? yesterday.contributionCount > 0 : false);`

Fixes #220 

## Pillar

- [ x] 🛠️ Other (Bug fix, refactoring, docs)

- [x ] I have read the `CONTRIBUTING.md` file.
- [x ] I have tested these changes locally (`localhost:3000/api/streak?user=YOUR_USERNAME`).
- [x ] I have run `npm run format` and `npm run lint` locally and resolved all errors (CI will fail otherwise).
- [x ] My commits follow the Conventional Commits format (e.g., `feat(themes): ...`, `fix(calculate): ...`).
- [x ] I have updated `README.md` if I added a new theme or URL parameter.
- [ x] I have started the repo.
- [ x] I have made sure that i have only one commit to merge in this PR.
- [ x] The SVG output matches the CommitPulse "premium quality" aesthetic standard (no raw elements, smooth animations, correct fonts).
- [ x] (Recommended) I joined the CommitPulse Discord community for contributor discussions, mentorship, and faster PR support.

### DIFF
--- a/lib/calculate.test.ts
+++ b/lib/calculate.test.ts
@@ -164,4 +164,20 @@ describe('calculateStreak', () => {
     expect(result.totalContributions).toBe(1);
     expect(result.longestStreak).toBe(1);
   });
+
+  it('handles a single inactive day safely (0 contributions)', () => {
+    const calendar = buildCalendar([0]);
+    expect(() => calculateStreak(calendar)).not.toThrow();
+    const result = calculateStreak(calendar);
+    expect(result.currentStreak).toBe(0);
+    expect(result.longestStreak).toBe(0);
+  });
+
+  it('handles an empty contribution calendar safely without crashing', () => {
+    const calendar = buildCalendar([]);
+    expect(() => calculateStreak(calendar)).not.toThrow();
+    const result = calculateStreak(calendar);
+    expect(result.currentStreak).toBe(0);
+    expect(result.longestStreak).toBe(0);
+  });
 });
--- a/lib/calculate.ts
+++ b/lib/calculate.ts
@@ -22,13 +22,22 @@ export function calculateStreak(calendar: ContributionCalendar): StreakStats {
   // 2. Calculate Current Streak (Backwards loop with Grace Period)
   // We look at the very last day in the array (Today in UTC)
   const todayIndex = days.length - 1;
+  if (todayIndex < 0) {
+    return {
+      currentStreak: 0,
+      longestStreak: 0,
+      totalContributions: calendar.totalContributions,
+    };
+  }
+
   const today = days[todayIndex];
-  const yesterday = days[todayIndex - 1];
+  const yesterday = todayIndex > 0 ? days[todayIndex - 1] : null;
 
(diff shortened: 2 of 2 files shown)


Aamod007 (approved):
Hey @Animesh-86, nice catch. This crash would've hit new accounts for sure, and your fix handles both edge cases cleanly. The early return and null check are exactly right, plus the new tests lock it in. All good locally lint, typecheck, and 147 tests. Ready to merge!

PR #812, opened 2026-05-27
fix(api): guard malformed GraphQL error responses

### LINKED ISSUE #807: [Bug]: GraphQL error handling assumes errors[0].message always exists
While reading through the GitHub GraphQL client, I noticed one brittle error path that can throw the wrong exception before the app gets a chance to return a clean API error.

Line: `lib/github.ts:267-268`

The code currently does this:

```ts
if (data.errors) {
  throw new Error(data.errors[0].message);
}
```

That works only when GitHub returns an `errors` array with at least one object and that object has a `message` string. If the response is malformed, proxied, mocked in tests, or otherwise shaped like `{ errors: [] }` / `{ errors: [{}] }`, this line can throw a TypeError instead of a controlled GitHub API error.

A quick way to verify the bug is to mock the GraphQL response in `lib/github.test.ts` with one of these payloads:

```json
{ "errors": [] }
```

or:

```json
{ "errors": [{}] }
```

Current behavior: the client dereferences `data.errors[0].message` without checking that index `0` exists or that `message` is a string.

Expected behavior: GraphQL errors should be normalized into a predictable message, even when the response shape is incomplete. For example, use the first string message if present, otherwise fall back to something like `GitHub GraphQL API returned an unknown error`.

Impact: this makes upstream API failures harder to diagnose and can turn a handled GitHub error into an unrelated runtime exception. Since this function powers the public streak SVG route, a malformed error response can surface as a generic broken badge instead of a useful controlled failure.

Suggested fix: guard the array before reading from it:

```ts
if (Array.isArray(data.errors) && data.errors.length > 0) {
  const message = typeof data.errors[0]?.message === "string"
    ? data.errors[0].message
    : "GitHub GraphQL API returned an unknown error";
  throw new Error(message);
}
```

It would also be worth adding tests for empty `errors`, missing `message`, and a normal GitHub error message.

Labels: `bug`, `api`, `GSSoC 2026`, `good first issue`

Opening this as part of GSSoC 2026.

eshaanag:
## Description

Fixes #807

## Pillar

## What changed
The GitHub GraphQL client assumed every GraphQL error response had a non-empty `errors` array with a `message` string at index `0`.

That meant responses like:

```json
{ "errors": [] }
```

or:

```json
{ "errors": [{}] }
```

could throw a runtime `TypeError` instead of a controlled API error. I added a small guard that normalizes malformed GraphQL error payloads to a stable fallback message while preserving the existing behavior for normal GitHub errors such as `Bad credentials`.

I also made the response `data` field optional in the local response type and changed the user lookup to use optional chaining so malformed responses do not crash while checking `data.user`.

## Visual Preview
N/A — this is API error handling and test coverage only.

## Testing
```bash
npm run format
npm run test -- lib/github.test.ts
npm run lint
npm run typecheck
npm run test
npm run build
```

### DIFF
--- a/lib/github.test.ts
+++ b/lib/github.test.ts
@@ -135,8 +135,32 @@ describe('fetchGitHubContributions', () => {
       })
     );
 
-    // Only the first error surfaces — the source always reads errors[0].
+    // Preserve the existing behavior of surfacing the first GitHub error message.
     await expect(fetchGitHubContributions('octocat')).rejects.toThrow('Bad credentials');
+  });
+
+  it('throws a stable fallback when GraphQL returns an empty errors array', async () => {
+    vi.mocked(fetch).mockResolvedValue(
+      mockResponse({
+        errors: [],
+      })
+    );
+
+    await expect(fetchGitHubContributions('octocat')).rejects.toThrow(
+      'GitHub GraphQL API returned an unknown error'
+    );
+  });
+
+  it('throws a stable fallback when the first GraphQL error has no message', async () => {
+    vi.mocked(fetch).mockResolvedValue(
+      mockResponse({
+        errors: [{}],
+      })
+    );
+
+    await expect(fetchGitHubContributions('octocat')).rejects.toThrow(
+      'GitHub GraphQL API returned an unknown error'
+    );
   });
 
   it('throws a descriptive "user not found" error when the username does not exist on GitHub', async () => {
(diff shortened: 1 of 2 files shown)


eshaanag:
Hi @maintainer, this PR was raised as part of GSSoC 2026 contribution.
Could you please add the appropriate GSSoC label if applicable? Thanks!

PR #1129, opened 2026-05-29
fix(svg): escape tower tooltip titles

### LINKED ISSUE #1128: [Bug]: Tower tooltip text is not XML-escaped in SVG output
I noticed this while checking the SVG renderer escape paths. User-facing strings like the username already go through `escapeXML`, but the per-tower tooltip text is inserted directly into `<title>` in both render paths.

Line: `lib/svg/generator.ts:193` and `lib/svg/generator.ts:401`

Problem: the renderer currently outputs:

```ts
<title>${t.tooltip}</title>
```

`t.tooltip` is built from calendar data in `lib/svg/layout.ts`, and the SVG renderer treats it as already-safe XML. If a malformed, mocked, or unexpected contribution day date contains characters like `&`, `<`, `>`, quotes, etc., the generated SVG title can become invalid XML. The rest of the file already has an `escapeXML` helper, so this path is inconsistent with the renderer's own escaping approach.

A quick way to verify this is to add a generator test with a contribution day date like:

```ts
{ contributionCount: 3, date: '2024-06-12 & <bad>' }
```

Current behavior: the output contains the raw tooltip text inside `<title>`, which can break the SVG markup.

Expected behavior: tooltip text should be XML-escaped before it is written into the SVG, just like usernames and metadata are.

Impact: this is a public SVG renderer, so one unsafe text path can produce malformed badge output and make edge-case data harder to debug. It is especially easy to miss because normal GitHub dates are safe, but the renderer itself should still guard the XML boundary.

Suggested fix: wrap `t.tooltip` with the existing `escapeXML` helper in both tower renderers and add tests for default and auto-theme SVG output using a tooltip/date string with XML-reserved characters.

Labels: `bug`, `frontend`, `GSSoC 2026`, `good first issue`

eshaanag:
## Description

Fixes #1128

## Pillar

- [x] 📐 Pillar 2 — Geometric SVG Improvement

## What changed

Tower tooltip text was being inserted directly into SVG `<title>` nodes in both the normal renderer and the auto-theme renderer. That meant XML-reserved characters from tooltip content could produce malformed SVG output.

This PR reuses the existing `escapeXML` helper for tower tooltip titles in both render paths:

- `lib/svg/generator.ts` default tower renderer
- `lib/svg/generator.ts` auto-theme tower renderer

I also added regression coverage for both renderers using a contribution day date containing `&` and `<` to verify the generated tooltip title is escaped correctly.

## Visual Preview

N/A — this is an SVG metadata escaping fix and does not change the visible badge design.

- [x] I have tested these changes locally (`npm run test -- lib/svg/generator.test.ts`, `npm run test`, `npm run test:coverage`).

### DIFF
--- a/lib/svg/generator.test.ts
+++ b/lib/svg/generator.test.ts
@@ -226,6 +226,25 @@ describe('generateSVG', () => {
     expect(svg).toMatch(/style="animation-delay: \d+\.\d+s;"/);
   });
 
+  it('escapes XML-reserved characters in tower tooltip titles', () => {
+    const calendarWithUnsafeDate = {
+      weeks: [
+        {
+          contributionDays: [{ contributionCount: 3, date: '2024-06-12 & <bad>' }],
+        },
+      ],
+    } as ContributionCalendar;
+
+    const svg = generateSVG(
+      mockStats,
+      { user: 'avi' } as unknown as BadgeParams,
+      calendarWithUnsafeDate
+    );
+
+    expect(svg).toContain('<title>TODAY: 2024-06-12 &amp; &lt;bad&gt;: 3 contributions</title>');
+    expect(svg).not.toContain('<title>TODAY: 2024-06-12 & <bad>: 3 contributions</title>');
+  });
+
   it('includes reduced-motion CSS for the scan line in the main SVG output', () => {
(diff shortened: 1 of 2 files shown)


eshaanag:
Raised as part of GSSoC 2026 for #1128. I do not have permission to add labels directly, so could a maintainer please add the appropriate GSSoC/bug labels if this PR looks valid?

PR #1244, opened 2026-05-29
fix(api): encode GitHub REST username paths

### LINKED ISSUE #1242: [Bug]: GitHub REST username is inserted into URL paths without encoding
I noticed this while checking the REST side of the GitHub client after testing a few malformed dashboard usernames. The GraphQL path sends the username as a variable, but the REST profile/repo helpers build the URL path directly from the raw `username` string.

Line: `lib/github.ts:310`, `lib/github.ts:338`, and `lib/github.ts:362`

Problem: the code currently does this:

```ts
`${GITHUB_REST_URL}/users/${username}`
`${GITHUB_REST_URL}/users/${username}/repos?per_page=100&page=1&sort=pushed`
```

If `username` contains a slash, `?`, `#`, or other path/query characters, the request URL no longer points to the intended GitHub user resource. For example, `octo/cat` becomes `/users/octo/cat` instead of treating the whole value as one path segment.

A quick way to verify it is to add a test around `fetchUserProfile('octo/cat')` or `fetchUserRepos('octo/cat')` and inspect the mocked `fetch` URL. Today it receives the raw slash in the path.

Current behavior: raw username input is interpolated directly into GitHub REST URLs.

Expected behavior: the username should be encoded before being used as a REST path segment, e.g. `encodeURIComponent(username)`, so malformed input cannot reshape the URL path or query string.

Impact: this makes invalid or unexpected dashboard username input harder to reason about and can send requests to the wrong GitHub endpoint. It also makes the REST client inconsistent with the GraphQL client, which does not manually splice the username into a URL.

Suggested fix: compute an `encodedUsername` inside the REST helpers and use it for the profile URL and all repo pagination URLs. Add tests that assert usernames with `/` are sent as `%2F` in both profile and repo fetches.

Labels: `bug`, `api`, `security`, `GSSoC 2026`, `good first issue`

eshaanag:
## Description

Fixes #1242

## Pillar

## What changed

The GitHub REST helpers were inserting the raw `username` value directly into `/users/${username}` URL paths. That works for normal GitHub usernames, but malformed input containing path/query characters can reshape the REST URL instead of being treated as one path segment.

This PR encodes the username with `encodeURIComponent` before building REST profile and repo URLs:

- `fetchUserProfile`
- `fetchUserRepos`
- repo pagination requests

I also added regression tests for usernames containing `/` so the mocked REST URLs now assert `%2F` is used in both profile and repo fetches.

## Visual Preview

N/A — this is an API client URL handling fix with no visual change.

- [x] I have tested these changes locally (`npm run test -- lib/github.test.ts`, `npm run test`, `npm run test:coverage`).

### DIFF
--- a/lib/github.test.ts
+++ b/lib/github.test.ts
@@ -84,7 +84,9 @@ describe('fetchWithRetry', () => {
     const abortListener = addListenerSpy.mock.calls.find(([event]) => event === 'abort')?.[1];
 
     expect(abortListener).toEqual(expect.any(Function));
-    expect(addListenerSpy).toHaveBeenCalledWith('abort', abortListener, { once: true });
+    expect(addListenerSpy).toHaveBeenCalledWith('abort', abortListener, {
+      once: true,
+    });
     expect(removeListenerSpy).toHaveBeenCalledWith('abort', abortListener);
   });
 
@@ -102,7 +104,9 @@ describe('fetchWithRetry', () => {
         })
     );
 
-    const request = fetchWithRetry('https://api.github.com/test', { signal: controller.signal });
+    const request = fetchWithRetry('https://api.github.com/test', {
+      signal: controller.signal,
+    });
 
     controller.abort();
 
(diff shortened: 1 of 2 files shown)


eshaanag:
Raised as part of GSSoC 2026 for #1242. I do not have permission to add labels directly, so could a maintainer please add the appropriate GSSoC/bug labels if this PR looks valid?

PR #1284, opened 2026-05-29
fix: prevent redirect() from being swallowed by try-catch in dashboard [username] page

### LINKED ISSUE #1280: CRITICAL: redirect() swallowed by try-catch in dashboard [username] page -- Organization redirect shows 404 instead
## Description

In `app/(root)/dashboard/[username]/page.tsx`, the `redirect()` call from `next/navigation` is placed inside a nested try-catch block. Since Next.js `redirect()` works by throwing a special `RedirectError`, the catch block intercepts it and returns `notFound()` instead. Users visiting `/dashboard/<orgname>` for a GitHub Organization see a 404 page instead of being redirected to `/dashboard/org/<orgname>`.

## Issue Cause

Next.js `redirect()` from `next/navigation` throws a `RedirectError` with a `NEXT_REDIRECT` digest. The framework normally catches this at the render level to perform the HTTP redirect. However, when called inside a try-catch, JavaScript catches the `RedirectError` first and the inner catch block returns `notFound()` -- so the redirect never reaches the framework.

## How I Would Fix It

Move the `redirect()` call to outside the try-catch block, or restructure the logic so the catch block does not swallow the `RedirectError`. The fix is to check the fallback profile type before entering the catch, or re-throw the redirect error explicitly.

## Additional Context

- This is a guaranteed failure path for every Organization URL visited directly
- The code already has a smart redirect intended to handle Organizations -- but it is completely broken
- The bug exists at lines 79-80 where `redirect()` is called inside try {} at line 76
- Once fixed, visiting `/dashboard/tailwindcss` (an org) should redirect to `/dashboard/org/tailwindcss` instead of showing 404

AnushKamble:
## Description

Fixes #1280

The Organization smart redirect in the dashboard [username] page was completely broken. When a user visits `/dashboard/<orgname>` for a GitHub Organization, they saw a 404 page instead of being redirected to `/dashboard/org/<orgname>`. This happened because Next.js `redirect()` from `next/navigation` throws a `RedirectError` internally, and the surrounding try-catch block was intercepting it before the framework could handle it.

## Root Cause

The `redirect()` call was nested inside a try-catch block. Since Next.js implements redirects by throwing a special error, the catch block caught it and returned `notFound()` instead, preventing the redirect from ever reaching the framework.

## Changes

Extracted the `redirect()` call outside the nested try-catch. The `fetchUserProfile` call remains protected by try-catch for network errors, but the `redirect()` now propagates to the framework properly.

## Pillar

- [ ] Pillar 1 - New Theme Design
- [ ] Pillar 2 - Geometric SVG Improvement
- [ ] Pillar 3 - Timezone Logic Optimization
- [X] Other (Bug fix, refactoring, docs)

## Visual Preview

N/A - Server-side redirect fix, no visual changes.

- [X] I have read the `CONTRIBUTING.md` file.
- [X] I have tested these changes locally.
- [X] I have run `npm run format` and `npm run lint` locally and resolved all errors (CI will fail otherwise).
- [X] My commits follow the Conventional Commits format (e.g., `feat(themes): ...`, `fix(calculate): ...`).
- [X] I have started the repo.
- [X] I have made sure that i have only one commit to merge in this PR.
- [ ] The SVG output matches the CommitPulse premium quality aesthetic standard (no raw elements, smooth animations, correct fonts).

### DIFF
--- a/app/(root)/dashboard/[username]/page.tsx
+++ b/app/(root)/dashboard/[username]/page.tsx
@@ -73,14 +73,14 @@ export default async function DashboardPage({
   } catch (error) {
     if (error instanceof Error && error.message.includes('not found')) {
       // Smart Redirect: If the GraphQL "user" query fails, check if it's actually an Organization
+      let fallbackProfile;
       try {
-        const fallbackProfile = await fetchUserProfile(username, { bypassCache });
-        if (fallbackProfile.type === 'Organization') {
-          redirect(`/dashboard/org/${username}`);
-        }
+        fallbackProfile = await fetchUserProfile(username, { bypassCache });
       } catch {
-        // If it's truly neither a user nor an org, show 404
         return notFound();
+      }
+      if (fallbackProfile.type === 'Organization') {
+        redirect(`/dashboard/org/${username}`);
       }
       return notFound();
     }


AnushKamble:
Hi @jhasourav07, could you please add the relevant labels (like `critical`, `bug`, `performance`) to this PR if it qualifies? This fixes the Organization smart redirect that was completely broken -- users were seeing a 404 instead of being redirected to the org dashboard. Thanks!

PR #1589, opened 2026-05-29
fix(api): dedupe concurrent GitHub cache misses

### LINKED ISSUE #1588: [Bug]: Concurrent cold cache misses duplicate GitHub API requests
I noticed this while reading through the GitHub client cache paths. The caches work after a response has completed, but there is no in-flight request deduplication while the first request is still pending.

Line: `lib/github.ts:254-258`, `lib/github.ts:359-364`, and `lib/github.ts:393-398`

Problem: each fetch method checks the cache, sees a miss, and immediately starts a GitHub request. The cache is only written later after the response finishes:

```ts
const cached = await contributionsCache.get(key);
if (cached) return cached;

const res = await fetchWithRetry(...)
```

That means two or more concurrent requests for the same username/year can all miss the cache at the same time and all hit GitHub independently. The first completed response will populate the cache, but by then the duplicate upstream requests have already been sent.

A realistic way to verify this is to mock `fetch` in `lib/github.test.ts`, add a small delay before resolving the first GraphQL response, and call:

```ts
await Promise.all([
  fetchGitHubContributions('octocat'),
  fetchGitHubContributions('octocat'),
  fetchGitHubContributions('octocat'),
]);
```

Current behavior: the mocked GitHub fetch is called three times on a cold cache.

Expected behavior: identical concurrent cache misses should share the same pending promise, so GitHub is called once and all callers receive the same result.

Impact: this can amplify traffic to GitHub during README badge bursts, dashboard reloads, or deploy cold starts. Since the app already has cache keys and retry/rate-limit handling, request coalescing would reduce avoidable API calls and lower the chance of hitting rate limits.

Suggested fix: keep a small in-flight `Map<string, Promise<T>>` per data type, keyed the same way as the cache. On a cold miss, store the pending promise before starting the fetch, return it to matching callers, and remove it in `finally`. `bypassCache` / `refresh=true` should continue to skip this behavior so forced refreshes remain explicit.

Labels: `bug`, `api`, `performance`, `GSSoC 2026`

eshaanag:
## Description

Fixes #1588

## Pillar

## What changed

The GitHub client already cached completed responses, but concurrent cold misses for the same cache key could still trigger duplicate upstream GitHub requests before the first response populated the cache.

This PR adds small in-flight request maps for contribution, profile, and repo fetches. Normal cacheable requests now share a pending promise for the same key, and the pending entry is removed once the request settles.

`bypassCache` is intentionally unchanged: refresh requests still skip both the completed cache and in-flight deduping so an explicit refresh remains a real fresh request.

I also added regression tests for concurrent cold misses across all three GitHub data paths.

This is a GSSoC 2026 performance/API bug fix contribution. Please add the relevant GSSoC labels if they are missing.

## Visual Preview

N/A — this is an API/client caching behavior change.

## Local checks

```bash
npm run format:check
npm run lint
npm run test -- lib/github.test.ts
```

- [x] My commit follows the Conventional Commits format.

### DIFF
--- a/lib/github.test.ts
+++ b/lib/github.test.ts
@@ -781,6 +781,37 @@ describe('GitHub API cache behavior', () => {
     expect(fetch).toHaveBeenCalledTimes(1);
   });
 
+  it('dedupes concurrent contribution requests for the same cold cache key', async () => {
+    let resolveFetch!: (response: Response) => void;
+    vi.mocked(fetch).mockImplementation(
+      () =>
+        new Promise<Response>((resolve) => {
+          resolveFetch = resolve;
+        })
+    );
+
+    const requests = Promise.all([
+      fetchGitHubContributions('octocat'),
+      fetchGitHubContributions('octocat'),
+      fetchGitHubContributions('octocat'),
+    ]);
+
+    await vi.waitFor(() => expect(fetch).toHaveBeenCalledTimes(1));
+
+    resolveFetch(
+      mockResponse({
+        data: {
+          user: {
+            contributionsCollection: { contributionCalendar: mockCalendar },
+          },
+        },
+      })
+    );
+
+    const results = await requests;
+    expect(results.map((result) => result.totalContributions)).toEqual([42, 42, 42]);
+  });
+
   it('refresh bypass: bypassCache=true forces a fresh fetch', async () => {
     vi.mocked(fetch).mockImplementation(async () =>
       mockResponse({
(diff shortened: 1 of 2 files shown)


eshaanag:
Updated the branch after the CI failure. The repo checks are passing now: format/lint/typecheck/test, production build, assignment check, and conflict check are all green. The only remaining failure is the Vercel authorization gate, which looks maintainer-side.

PR #2096, opened 2026-05-31
fix: apply grace parameter to longest streak calculation

### LINKED ISSUE #2069: bug: Longest streak calculation ignores the grace parameter
## Description

The current streak calculation respects the `grace` parameter, but the longest streak calculation does not. This creates inconsistent behavior.

### Problem

In `lib/calculate.ts`:

- **Current streak** (lines 62-81): Respects the `grace` parameter, allowing gaps of up to `grace` days without breaking the streak.
- **Longest streak** (lines 39-47): Uses a simple consecutive-days loop with no grace period consideration.

This means `longestStreak` and `currentStreak` use different definitions of what constitutes a "streak."

### Example

If a user sets `grace=2`, they would expect their longest streak to also tolerate 2-day gaps. However, only the current streak respects this — the longest streak still requires strict consecutive days.

### Expected behavior

The longest streak calculation should apply the same `grace` period logic as the current streak calculation, so both use a consistent definition of "streak."

### Files

- `lib/calculate.ts` (lines 39-47 for longest streak, lines 62-81 for current streak)

Tranquil666:
## Description

Fixes #2069

The longest streak calculation used strict consecutive-day logic while the current streak respected the `grace` parameter. This created inconsistent behavior — users setting `grace=2` expected both streaks to tolerate gaps, but only the current streak did.

The new algorithm tracks `gapDays` (consecutive zero-contribution days). When `gapDays > grace`, the streak resets. Only contribution days count toward the streak total.

### Changes
- **`lib/calculate.ts`** — Replaced strict consecutive-day loop with grace-aware longest streak algorithm
- **`lib/calculate.test.ts`** — Updated 3 test expectations + added 1 new test for grace=2 longest streak

## Pillar

- [x] 🕐 Pillar 3 — Timezone Logic Optimization

## Visual Preview

No visual changes — this is a calculation logic fix. The SVG badge will now show correct longest streak values when the `grace` parameter is set.

### DIFF
--- a/lib/calculate.test.ts
+++ b/lib/calculate.test.ts
@@ -169,7 +169,9 @@ describe('calculateStreak', () => {
 
     const result = calculateStreak(calendar);
 
-    expect(result.longestStreak).toBe(7);
+    // With grace=1 (default), the single zero-day gap between the 7-day and
+    // 6-day runs is within tolerance, so they merge into a single 13-day streak.
+    expect(result.longestStreak).toBe(13);
     expect(result.currentStreak).toBe(6);
     expect(result.totalContributions).toBe(13);
   });
@@ -498,7 +500,9 @@ describe('calculateStreak', () => {
       new Date('2024-03-01T12:00:00Z')
     );
     expect(resultLeapGap.currentStreak).toBe(1);
-    expect(resultLeapGap.longestStreak).toBe(1);
+    // With grace=1 (default), the single-day gap on Feb 29 is within tolerance,
+    // so longest streak merges Feb 28 + Mar 1 = 2 contribution days.
+    expect(resultLeapGap.longestStreak).toBe(2);
   });
 
   it('handles leap years vs non-leap years Feb 28 to Mar 1 timeline and asserts correct current/longest streaks', () => {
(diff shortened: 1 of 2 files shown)


Aamod007 (changes requested):
Thanks for the PR @Tranquil666.

However, I cannot approve this change because it relies on a fundamental misunderstanding of what the grace parameter does. 

The grace parameter is not a historical hole-puncher for missing days. It exists solely to provide timezone/delay tolerance for the *current edge of the timeline* (e.g., if it is morning in your timezone and you haven't committed yet today, grace=1 prevents your current streak from dropping to 0 by tolerating a 0 on the *current* day).

If you look closely at the currentStreak calculation in lib/calculate.ts, you'll see that it **does not** tolerate 0-contribution days in the middle of a streak—it only tolerates them at the very end (i >= todayIndex - grace). 

If we apply this PR's logic to longestStreak, it will allow users to bridge 1-day gaps anywhere in their history infinitely. A user could commit every other day for a year and this PR would calculate that as a 365-day streak. That completely breaks the definition of a consecutive streak.

Please close this PR as the current logic is actually correct!

PR #2196, opened 2026-05-31
fix(cache): fix ttl cache update method

### LINKED ISSUE #1879: bug: TTLCache.update() does not remove expired entries, causing stale keys to remain in memory
### Description

Summary

The update() method in lib/cache.ts correctly returns false when a cache entry has expired, but it does not delete the expired entry from the internal store.

This results in expired keys remaining in memory until a later cleanup sweep runs, leading to unnecessary memory usage and inconsistent behavior compared to get() and has(), which immediately remove expired entries.

Location

lib/cache.ts

update(key: string, value: T): boolean {
  const hit = this.store.get(key);
  if (!hit || Date.now() > hit.expiresAt) return false;
  hit.value = value;
  return true;
}

### Steps to Reproduce

1. Create a cache entry with a very short TTL.
2. Wait until the entry expires.
3. Call update() on the expired key.
4. Observe that:
update() returns false.
5. The expired entry remains in store until a cleanup sweep occurs.

### Expected Behavior

Expired entries should be removed immediately when detected inside update(), matching the behavior of get() and has()

vbarik317-droid:
## Summary

Fixes an issue where `TTLCache.update()` returned `false` for expired entries but did not remove them from the internal cache store.

## Changes Made

* Added immediate cleanup of expired entries inside `update()`.
* Deletes the expired key from `this.store` before returning `false`.
* Keeps behavior consistent with `get()` and `has()`, which already remove expired entries when encountered.

## Expected Result

Expired cache entries are no longer left in memory after calling `update()`, reducing stale data retention and ensuring consistent cache behavior.

Fixes #1879


Aamod007:
@vbarik317-droid work on pipleine tests

Aamod007:
@vbarik317-droid resolve conflicts

Aamod007 (approved):

vbarik317-droid:
@JhaSourav07  "Fixed the Prettier formatting conflict in lib/cache.ts and resolved the merge conflict. CI should pass now."

Aamod007 (approved):

PR #2429, opened 2026-06-01
feat(viewer): add dark/light mode toggle

### LINKED ISSUE #2425: feat: add dark/light mode toggle for the contribution graph embed preview page

I'm always frustrated when I visit the CommitPulse preview page to check how my graph looks — the page itself has a fixed background, which makes it hard to accurately visualize how the SVG embed will actually appear in a dark-themed or light-themed README. There's no way to quickly toggle the background context without manually editing the URL or opening it in a separate tool.

Add a dark/light mode toggle button on the contribution graph preview page. This would allow users to instantly switch the page background between dark and light, so they can see exactly how their isometric SVG will look in both GitHub README contexts — without changing any URL parameters.

Manually appending a `bg` color parameter to the URL to test different backgrounds. However, this is cumbersome and not intuitive for new users. Another alternative is opening the embed in a custom HTML page, but that's too much friction for a quick visual check.

janievinod:
Closes #2425

## What's changed
- Added `isDarkBg` state to `InteractiveViewer` component
- Added a toggle button (☀️ Light / 🌙 Dark) in the top-right corner of the viewer
- Background switches between `bg-zinc-950` (dark) and `bg-white` (light)

## Why
Users couldn't visualize how their SVG embed would look in both dark and light README contexts without manually editing URL params.

## Testing
- Toggle button appears on the InteractiveViewer
- Clicking switches background between dark and light
- Button click does not trigger drag/pan behavior (`stopPropagation`)


Aamod007 (changes requested):
Hey! It looks like this PR broke the tests in `components/InteractiveViewer.test.tsx` (14 failing tests!).

The issue is that the tests use `viewerContainer.children[1]` to find the inner content wrapper, but by adding the new `<button>` as the very first child of the container, you shifted all the DOM children by one index! The tests are now asserting against the parallax background instead of the content wrapper.

You have two options:
1. Place the `<button>` at the *bottom* of the container (after the content div).
2. (Preferred) Update `components/InteractiveViewer.test.tsx` to query for the content div differently (e.g. adding a `data-testid` to the content wrapper and querying by that instead of using fragile child indices).

Let me know when you've pushed a fix!

Aamod007 (changes requested):
Hey! It looks like adding the `<button>` inside the container shifted the DOM child elements, which broke 14 tests in `components/InteractiveViewer.test.tsx`. The tests were assuming a specific index (like `viewerContainer.children[1]`) for the content element.

Could you please update the tests to select the correct child element (or use a `data-testid`), and run `npx vitest run components/InteractiveViewer.test.tsx` locally to ensure everything passes?

Also, I've updated the PR title to `feat(viewer): add dark/light mode toggle` to reflect your changes since the previous one was incorrect!

Let me know when you push a fix.

Aamod007 (changes requested):
The build fails because there are unresolved merge conflict markers in \lib/svg/generator.ts\, \lib/validations.ts\, and \	ypes/index.ts\. Please resolve these conflicts.

PR #3412, opened 2026-06-03
fix(og): stop bypassing cache on every OG image request

### LINKED ISSUE #3398: bug: OG image endpoint always bypasses cache, firing a fresh GitHub GraphQL request on every crawl
## Description

In `app/api/og/route.tsx`, the GitHub contributions fetch is hardcoded to `{ bypassCache: true }`:

```ts
const calendar = await fetchGitHubContributions(user, { bypassCache: true });
```

The `/api/og` endpoint is embedded as a `<meta>` tag on every user dashboard page and is hit by link-preview bots (Slack, Twitter, Discord, iMessage, etc.) on every share. Because `bypassCache` is always `true`, each of those crawls fires a fresh GitHub GraphQL API request regardless of how recently the same user's data was fetched. This silently ignores the caching infrastructure that every other route (`/api/stats`, `/api/streak`, `/api/wrapped`) uses correctly.

The response also carries `Cache-Control: public, max-age=3600` — meaning CDN and browser layers believe the image is cached for an hour, while the underlying GitHub data fetch is never cached at all. The two layers are inconsistent.

Additionally, `ogParamsSchema` has no `refresh` field, so there is no way for callers to opt into a fresh fetch when they actually need one.

## Steps to Reproduce

1. Share any CommitPulse dashboard URL (e.g. in Slack or Discord).
2. Observe in server logs that each unfurl fires a new outbound GitHub GraphQL request.
3. Share the same URL repeatedly — quota is consumed every time with no cache hit.

## Expected Behavior

The GitHub fetch should use the shared cache by default (`bypassCache: false`), matching the behavior of all other API routes. A `?refresh=true` parameter should allow callers to explicitly bypass the cache when needed, with the `Cache-Control` response header updated to `no-cache, no-store, must-revalidate` to stay consistent.

## Actual Behavior

- Every request to `/api/og` fires a fresh GitHub GraphQL call regardless of cache state.
- The `Cache-Control: public, max-age=3600` response header is misleading — the image is served from CDN cache but the data behind it was always freshly fetched.
- No `?refresh=true` escape hatch exists on this endpoint, unlike every other API route.

## Impact

> ⚠️ Under any meaningful share volume this will exhaust the unauthenticated GitHub API rate limit (60 req/hour) or burn through a configured token's quota, causing all stat endpoints across the app to start returning rate-limit errors.

## Files Affected

| File | Issue |
|---|---|
| `app/api/og/route.tsx` | Hardcoded `bypassCache: true`, hardcoded `Cache-Control`, uses `.parse()` instead of `.safeParse()` |
| `lib/validations.ts` | `ogParamsSchema` missing `refresh` field |

## Proposed Fix

**`lib/validations.ts`** — add `refresh` to `ogParamsSchema`:

```ts
refresh: z.string().optional().transform(toRefreshFlag),
```

**`app/api/og/route.tsx`** — switch to `safeParse`, honour the `refresh` flag, and align `Cache-Control` with it:

```ts
const parseResult = ogParamsSchema.safeParse(Object.fromEntries(searchParams.entries()));

if (!parseResult.success) {
  return new Response(JSON.stringify({ error: 'Invalid parameters', details: parseResult.error.flatten() }), {
    status: 400,
    headers: { 'Content-Type': 'application/json', 'Cache-Control': 'no-store' },
  });
}

const { user, theme, bg, text, accent, refresh } = parseResult.data;

// ...

const calendar = await fetchGitHubContributions(user, { bypassCache: refresh });

const cacheControl = refresh
  ? 'no-cache, no-store, must-revalidate'
  : 'public, max-age=3600, stale-while-revalidate=86400';
```

## Additional Notes

- `toRefreshFlag` is already defined in `lib/validations.ts` and used by every other schema — it just wasn't wired up here.
- This fix also resolves a secondary issue: `.parse()` throws an unhandled Zod exception on invalid params, crashing the route with a 500. Switching to `.safeParse()` returns a clean 400 instead.

Kokila-chandrakar:
## Description

Fixes a bug where the OG image endpoint was bypassing cache on every request, firing a fresh GitHub GraphQL call on each link-preview bot crawl (Slack, Discord, Twitter etc.) and silently exhausting API rate limit quota.

Closes #3398

## Pillar

Bug Fix

## Changes Made

- Switched `ogParamsSchema.parse()` to `safeParse()` with proper 400 error response
- Changed `bypassCache: true` to `bypassCache: refresh`
- Added `refresh` field to `ogParamsSchema`
- Aligned `Cache-Control` response header with actual fetch behaviour

## Checklist

- [x] My code follows the project style guidelines
- [x] I have performed a self-review of my code
- [x] I have run `npm run lint` and `npm run typecheck` locally
- [x] No new errors introduced
- [x] Related issue is linked above


Aamod007 (approved):
Great work! Code is clean and passes checks. LGTM!

PR #3514, opened 2026-06-03
fix(resume): verify uploaded file signature against declared type

### LINKED ISSUE #3513: bug: /api/student/resume/upload accepts files whose content does not match the declared type
## Description

`POST /api/student/resume/upload` (`app/api/student/resume/upload/route.ts`) decides whether to accept an upload based only on `ALLOWED_MIME_TYPES.includes(file.type)`. `file.type` is the Content-Type the client sends with the multipart part, so it is fully client controlled and can be set to any value. The actual bytes of the file are never inspected.

`parseResume` then calls `extractTextFromBuffer` (`lib/resume-parser.ts`), which ignores the type and runs `buffer.toString('utf-8')` over whatever was uploaded, feeding arbitrary content into the field extractors. So a file that is not a real PDF or DOCX (for example an HTML or plain text file) is accepted and processed as long as its declared type is in the allow list.

This is an input validation gap: the server should confirm that the actual file content matches an accepted document type before accepting and parsing it.

## Steps to Reproduce

1. Run the app locally (`npm run dev`).
2. Send an upload whose declared type is allowed but whose bytes are not a real PDF or DOCX, for example:
   `curl -F 'resume=@note.html;type=application/pdf' http://localhost:3000/api/student/resume/upload`
3. Observe the request is accepted (HTTP 200) and the HTML text is parsed into the resume fields, instead of being rejected.

## Expected Behavior

The endpoint verifies the file's leading bytes (magic number) against the declared type and rejects with `400` when they do not match. PDF must start with `%PDF-`, DOCX must be a ZIP container, and legacy DOC must be an OLE2 document.

## Screenshots / Logs

Not applicable. Reproducible with a single request.

## Environment

Other

Anexus5919:
## Description

Fixes #3513

Adds server-side file-signature (magic-byte) verification to the resume upload endpoint, so a file is only accepted when its actual bytes match its declared MIME type. Previously the client-controlled `file.type` header was the only gate, so any content (for example an HTML or text file) labelled with an allowed type was accepted and parsed.

Changes:

- Added `hasValidFileSignature(buffer, mimeType)` to `lib/resume-parser.ts`, which checks the file's leading bytes against the accepted types: PDF (`%PDF-`), DOCX / OOXML (ZIP `PK` signatures), and legacy DOC (OLE2).
- The upload route now rejects with `400` when the bytes do not match the declared type, before parsing.
- Implemented as a plain byte check with no new dependency, to fit the existing codebase.
- Added unit tests for the signature helper and route-level tests for the reject and accept paths.

Note on scope: for DOCX this confirms a valid ZIP container (the OOXML wrapper); it does not unzip to confirm the inner Word parts, which would require a heavier parser. It still rejects non-document content disguised with a document MIME type, which is the reported gap.

## Pillar

## Visual Preview

Not applicable. This is a backend API change with no visual output.

- [x] I have tested these changes locally (added unit tests for the signature helper and the upload route; the resume-parser suite and `tsc --noEmit` also pass).
- [ ] I have started the repo.
- [ ] The SVG output matches the CommitPulse "premium quality" aesthetic standard. (Not applicable, no SVG output in this change.)

### DIFF
--- a/app/api/student/resume/tests/upload.test.ts
+++ b/app/api/student/resume/tests/upload.test.ts
@@ -1,0 +1,62 @@
+import { describe, it, expect, beforeEach, vi } from 'vitest';
+import { POST } from '../upload/route';
+
+vi.mock('@/lib/rate-limit', () => {
+  class MockRateLimiter {
+    check = vi.fn().mockResolvedValue(true);
+  }
+
+  return {
+    RateLimiter: MockRateLimiter,
+  };
+});
+
+// Build the handler request directly from a real File/FormData so the file bytes are
+// preserved (the multipart transport itself is Next.js's responsibility, not this route's).
+function makeUploadRequest(content: string | number[], type: string, name = 'resume.pdf'): Request {
+  const data = typeof content === 'string' ? content : new Uint8Array(content);
+  const file = new File([data], name, { type });
+  const form = new FormData();
+  form.append('resume', file);
+
+  return {
+    headers: new Headers(),
+    formData: async () => form,
+  } as unknown as Request;
+}
+
+describe('POST /api/student/resume/upload', () => {
+  beforeEach(() => {
+    vi.clearAllMocks();
+  });
+
+  it('returns 400 for a disallowed mime type', async () => {
+    const response = await POST(makeUploadRequest('hello', 'text/html', 'note.html'));
+    const body = await response.json();
+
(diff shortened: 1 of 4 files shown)


Anexus5919:
@JhaSourav07 Kindly have a review on this pr. Thanks!

Aamod007 (approved):

PR #4017, opened 2026-06-05
Fix race condition in claim handler by scoping concurrency per-user

### LINKED ISSUE #3998: Race condition in /claim handler allows exceeding MAX_ASSIGNED_ISSUES cap
## Problem

The `/claim` handler at `.github/scripts/issue-management/claim-handler.js:68-80` has a TOCTOU (time-of-check-time-of-use) race condition. When two `/claim` commands are issued on **different issues** by the **same user** near-simultaneously, both workflow runs can read the same current assignment count, both see it below `MAX_ASSIGNED_ISSUES` (5), and both proceed to assign — resulting in 6+ active issues.

## Technical Root Cause

The race window exists because of two design issues:

### 1. Concurrency group is per-issue, not per-user

At `.github/workflows/issue-management.yml:8-9`:
```yaml
concurrency:
  group: issue-management-${{ github.event.issue.number }}
  cancel-in-progress: false
```

This serializes commands on the *same issue* but allows parallel execution on *different issues*. Two `/claim` commands on Issue A and Issue B run simultaneously.

### 2. Check-then-act without atomicity

At `.github/scripts/issue-management/claim-handler.js:68-80`:
```javascript
// Step 1 — READ (no lock)
const existingIssues = await findExistingAssignments(github, owner, repo, commenter, issueNumber);
// ... both workflow runs see count = 4 here ...

// Step 2 — WRITE (no CAS)
await github.rest.issues.addAssignees({
  owner, repo, issue_number: issueNumber, assignees: [commenter],
});
```

Between Step 1 and Step 2, another workflow instance for the same user on a different issue can also read `count = 4`, both pass the `MAX_ASSIGNED_ISSUES` check, and both assign. The `findExistingAssignments` function at lines 1-11 uses `listForRepo` with `assignee: username`, which is eventually consistent — it may not reflect the most recent `addAssignees` call from the other concurrent workflow.

## Affected Components

- `.github/scripts/issue-management/claim-handler.js:1-11` — `findExistingAssignments()` reads current assignments
- `.github/scripts/issue-management/claim-handler.js:68-80` — Check-then-assign without atomicity
- `.github/workflows/issue-management.yml:7-9` — Concurrency group scoped per-issue, not per-user
- `.github/scripts/issue-management/assign-handler.js:78-96` — Same pattern exists in `/assign` handler

## Reproduction

1. User A has 4 active assigned issues (under the limit of 5)
2. User A opens two browser tabs and comments `/claim` on Issue X and Issue Z within the same second
3. `issue-management.yml` triggers two workflow runs (one per issue), both executing in parallel
4. Both `findExistingAssignments` calls return 4
5. Both pass the `MAX_ASSIGNED_ISSUES < 5` check
6. Both call `addAssignees`
7. User A now has 6 assigned issues, exceeding the cap

## Expected Behavior

The `MAX_ASSIGNED_ISSUES` cap should be enforced atomically. Two concurrent claims should not both succeed when the user is at the limit.

## Actual Behavior

Parallel `/claim` commands on different issues bypass the cap. The count can drift arbitrarily high depending on concurrency.

## Production Impact

- **Anti-hoarding bypass:** The primary mechanism preventing issue hoarding in GSSoC is effectively broken under concurrent access
- **Unfair contention:** One contributor can occupy more than their fair share of open issues by racing their `/claim` commands
- **Blocked contributors:** Legitimate contributors may find all issue slots occupied by a single user who exploited the race

**Option A (Impact of change: low, effectiveness: high):** Change the concurrency group to scope on the commenter's username instead of the issue number:

```yaml
concurrency:
  group: issue-claim-${{ github.event.comment.user.login }}-${{ github.event.comment.id }}
  cancel-in-progress: true
```

This ensures all commands from the same user are serialized. Setting `cancel-in-progress: true` for the older run ensures only the latest claim attempt proceeds.

**Option B (More robust):** Replace the check-then-act with an atomic operation. Since GitHub's issue API doesn't support CAS (compare-and-swap) on assignees, implement a two-phase approach:
1. Before assigning, comment a "claim intent" marker on the issue
2. `findExistingAssignments` filters out issues that have an intent marker older than 5 minutes
3. Periodic cleanup of stale intent markers

**Option C:** Add a server-side lock using `DistributedCache` with Redis (already available at `lib/cache.ts`). Before the check, acquire a lock on `lock:claim:<username>`. Release after the assignment completes:
```javascript
const lockKey = `lock:${commenter}`;
const lockAcquired = await claimLock.tryAcquire(lockKey, 5000);
if (!lockAcquired) { /* retry or back off */ }
```

atul-upadhyay-7:
## What this fixes

The `/claim` handler at `.github/scripts/issue-management/claim-handler.js:68-80` has a TOCTOU race condition. When a user comments `/claim` on two different issues near-simultaneously, both workflow runs can read the same current assignment count, both see it below `MAX_ASSIGNED_ISSUES` (5), and both proceed to assign — allowing the user to exceed the cap. The same race exists in the `/assign` handler at `assign-handler.js:78-96`.

## Root cause

The concurrency group at `.github/workflows/issue-management.yml:7-9` was keyed on the issue number (`issue-management-${{ github.event.issue.number }}`). This serialized commands on the same issue but allowed parallel execution across different issues. Since both the check (`findExistingAssignments`) and the assignment (`addAssignees`) happen inside each workflow run with no mutual exclusion, two concurrent runs for the same user on different issues race through the check-then-act window.

## What changed

**`.github/workflows/issue-management.yml:8`** — Changed the concurrency group from per-issue to per-user:

```
issue-management-${{ github.event.issue.number }}
→
issue-cmd-${{ github.event.comment.user.login }}
```

`cancel-in-progress: false` is retained so no commands are dropped — concurrent runs from the same user are queued and processed sequentially.

## How to verify

1. Ensure the `MAX_ASSIGNED_ISSUES` limit is 5
2. As user A with 4 existing assignments, open two browser tabs
3. Comment `/claim` on issue X in tab 1 and `/claim` on issue Y in tab 2 simultaneously
4. Before the fix: both succeed, user A ends up with 6 assignments
5. After the fix: the second run is queued behind the first. When it starts, the count is already 5, so it rejects with the over-limit message

## Edge cases considered

- **Commands from different users**: These use different concurrency groups and still run in parallel — no regression for unrelated claims.
- **Same user, same issue**: Already serialized by the old concurrency group. The new group also serializes them (same user = same group).
- **`/unclaim` and `/claim` from the same user**: The `/unclaim` waits for the `/claim` to finish (or vice versa). This is correct — serial execution is better than a race.
- **`cancel-in-progress: true` not used**: No commands are dropped, which is important for `/unclaim` and `/addlabel` where cancellation would lose user intent.
- **Eventual consistency of GitHub API**: The second queued run starts after the first completes, giving the API time to propagate the assignment change before the next `listForRepo` call.

Closes #3998

### DIFF
--- a/.github/scripts/issue-management/assign-handler.js
+++ b/.github/scripts/issue-management/assign-handler.js
@@ -64,7 +64,13 @@ async function handleAssign({ github, context, username, hasWriteAccess }) {
     throw error;
   }
 
-  const currentAssignees = context.payload.issue.assignees.map((a) => a.login.toLowerCase());
+  // Re-fetch to avoid stale assignee data from the webhook payload
+  const { data: freshIssue } = await github.rest.issues.get({
+    owner,
+    repo,
+    issue_number: issueNumber,
+  });
+  const currentAssignees = freshIssue.assignees.map((a) => a.login.toLowerCase());
   if (currentAssignees.includes(username.toLowerCase())) {
     await github.rest.issues.createComment({
       owner,
--- a/.github/scripts/issue-management/claim-handler.js
+++ b/.github/scripts/issue-management/claim-handler.js
@@ -43,7 +43,13 @@ async function handleClaim({ github, context }) {
     return;
   }
 
-  const currentAssignees = context.payload.issue.assignees.map((a) => a.login.toLowerCase());
+  // Re-fetch to avoid stale assignee data from the webhook payload
+  const { data: freshIssue } = await github.rest.issues.get({
+    owner,
+    repo,
+    issue_number: issueNumber,
+  });
+  const currentAssignees = freshIssue.assignees.map((a) => a.login.toLowerCase());
 
   if (currentAssignees.length > 0) {
     if (currentAssignees.includes(commenter.toLowerCase())) {
--- a/.github/scripts/issue-management/stale-assignment.js
+++ b/.github/scripts/issue-management/stale-assignment.js
@@ -28,12 +28,24 @@ async function handleStaleAssignments({ github, context, core }) {
       const timeSinceUpdate = now.getTime() - updatedAt.getTime();
 
       if (timeSinceUpdate > TWO_DAYS_MS) {
+        const currentAssignees = issue.assignees.map((a) => a.login);
+
+        // Check if any open PRs reference this issue before unassigning
+        const { data: searchResult } = await github.rest.search.issuesAndPullRequests({
+          q: `"#${issue.number}" is:pr is:open repo:${owner}/${repo}`,
+        });
+
+        if (searchResult.total_count > 0) {
+          console.log(
+            `Issue #${issue.number} has open PR(s) referencing it. Skipping stale unassignment.`
+          );
+          continue;
+        }
+
         console.log(
           `Issue #${issue.number} has been inactive since ${issue.updated_at}. Removing assignees.`
         );
 
-        // 1. Remove all assignees
-        const currentAssignees = issue.assignees.map((a) => a.login);
         if (currentAssignees.length > 0) {
           await github.rest.issues.removeAssignees({
             owner,
--- a/.github/scripts/package-lock.json
+++ b/.github/scripts/package-lock.json
@@ -1,0 +1,271 @@
+{
+  "name": "commitpulse-issue-duplicate-detector",
+  "version": "1.0.0",
+  "lockfileVersion": 3,
+  "requires": true,
+  "packages": {
+    "": {
+      "name": "commitpulse-issue-duplicate-detector",
+      "version": "1.0.0",
+      "dependencies": {
+        "@actions/github": "^6.0.0",
+        "@google/generative-ai": "^0.11.0"
+      }
+    },
+    "node_modules/@actions/github": {
+      "version": "6.0.1",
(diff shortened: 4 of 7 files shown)


atul-upadhyay-7:
@Aamod007 can you review this

atul-upadhyay-7:
**Addressed copilot review:**

**Re: concurrency key concern** — The per-user scope is intentional. Including the issue number in the key (e.g., `issue-cmd-${{ issue.number }}-${{ comment.user.login }}`) would create different groups for the same user on different issues, allowing the TOCTOU race that is the primary bug here. Two simultaneous `/claim` calls from the same user on different issues would still race through the check-then-assign window. The per-user key serializes all commands from one user, which is the correct scope for enforcing `MAX_ASSIGNED_ISSUES`.

**Re: same-issue race between different users** — `claim-handler.js` and `assign-handler.js` now re-fetch the issue from the API (instead of reading the webhook payload) before checking current assignees. This mitigates the stale data problem and reduces the window for a same-issue double-assignment between different users. Combined with the per-user concurrency, this addresses both vectors.

Aamod007 (approved):

PR #4022, opened 2026-06-05
Add per-username write cooldown and GitHub validation to POST /api/notify

### LINKED ISSUE #3997: Authorization bypass: POST /api/notify accepts preferences for any GitHub username with no ownership verification
## Problem

The `POST /api/notify` endpoint at `app/api/notify/route.ts:100-112` performs an unconditional MongoDB upsert on the `Notification` collection keyed by `username`. The only access control is IP rate limiting (5 req/min per `lib/rate-limit.ts:189`). There is **no authentication, session check, or ownership proof** verifying the caller controls the GitHub account they're registering preferences for.

## Technical Root Cause

At `app/api/notify/route.ts:73-112`:

```typescript
const { username, email, frequency, preferences } = parsed.data;
// ... no auth/ownership check ...
const notification = await Notification.findOneAndUpdate(
  { username: username.toLowerCase() },
  { email: email.toLowerCase(), frequency, notifyOnCommit: ..., notifyOnStreak: ..., notifyOnMilestone: ... },
  { upsert: true, new: true }
);
```

The `notifyPostSchema` at `lib/validations.ts:536-566` validates field *format* (email regex, username regex, enum values) but never validates *ownership*. The flow is:

1. Parse body (Zod) — format check only
2. Rate limit check (IP-based, NOT user-based)
3. MongoDB upsert — no auth gate

An attacker can send `{ "username": "torvalds", "email": "attacker@evil.com", "frequency": "daily" }` and the server silently upserts the record with the attacker-controlled email.

## Affected Components

- `app/api/notify/route.ts:36-138` — POST handler with no auth middleware
- `lib/validations.ts:536-566` — `notifyPostSchema` only validates format, not ownership
- `lib/rate-limit.ts:189` — `notifyRateLimiter` is IP-based (5 req/min), trivially bypassed with rotating IPs or low-volume targeted attacks
- `models/Notification.ts` — schema with no owner verification field

## Reproduction

1. As an unauthenticated client, send:
   ```
   POST /api/notify
   Content-Type: application/json
   
   { "username": "jhasourav07", "email": "attacker-controlled@evil.com", "frequency": "daily", "preferences": { "notifyOnCommit": false, "notifyOnStreak": false, "notifyOnMilestone": false } }
   ```
2. The upsert succeeds — the legitimate user's notification preferences are now pointing to an attacker-controlled email
3. GET /api/notify?user=jhasourav07 confirms the email was changed (masked, but the tampered record is stored)

## Expected Behavior

The POST endpoint should verify that the caller either:
- Is authenticated with GitHub OAuth as the target user, OR
- Provides a proof-of-ownership token

## Actual Behavior

Any unauthenticated client can upsert notification preferences for any GitHub username. The only defense is IP rate limiting, which is trivially bypassed with rotating proxies or simply keeping requests under 5/min.

## Production Impact

- **Data tampering:** An attacker can set arbitrary emails for any GitHub user's notification preferences
- **Phishing:** If the notification system is later used to send emails, attackers control the destination address
- **Storage exhaustion:** Bulk upserts can inflate MongoDB storage cost
- **Cooldown bypass:** The 5-minute write cooldown in `track-user-protection.ts` is per-GitHub-username, meaning an attacker doing a targeted attack against specific usernames would bypass it

1. **Short term:** Add a write cooldown per-username (not just per-IP) to prevent rapid upserts against the same user. Use `DistributedCache` with a 5-minute TTL keyed on `username`.
2. **Medium term:** Implement GitHub OAuth verification. Use NextAuth.js or similar to obtain a GitHub session token. Verify the authenticated GitHub user's login matches `username` before allowing the upsert.
3. **Long term:** Move to a challenge-response flow. When a user registers, send a verification token to the email. The upsert should only commit after the token is verified.

Reference implementation pattern: `services/github/validate-user.ts` already has GitHub API user lookup — the notify POST could use the same service to verify the caller has a session for the target user.

atul-upadhyay-7:
## What this fixes

The `POST /api/notify` endpoint at `app/api/notify/route.ts:100-112` performed an unconditional MongoDB upsert keyed by `username` with no authentication, session check, or ownership proof. Any unauthenticated client could set notification preferences — including an arbitrary email address — for any GitHub username. The only defense was IP-based rate limiting (5 req/min per `lib/rate-limit.ts:189`), trivially bypassed with rotating proxies or by staying under the per-IP limit.

## Root cause

The POST handler in `app/api/notify/route.ts` accepted `{ username, email, frequency, preferences }` and ran `Notification.findOneAndUpdate` with `{ upsert: true }` directly. No auth middleware, no session check, and no verification that the caller controls the GitHub account they were registering preferences for. The Zod schema at `lib/validations.ts:536-566` only validated field format (email regex, username regex, enum values) — never ownership.

## What changed

- **`app/api/notify/route.ts`** — Added a per-username write cooldown using `DistributedCache` (5-minute TTL keyed on `username`), so an attacker can only upsert preferences for a given username once every five minutes. Added a GitHub user existence check via the existing `gitHubUserValidator` from `services/github/validate-user.ts`, so upserts for non-existent usernames are rejected. The upsert query now uses a pre-normalized `normalizedUsername` variable for consistency. Exported `resetNotifyWriteCache()` for test isolation.
- **`app/api/notify/route.test.ts`** — Mocked `gitHubUserValidator`, added `resetNotifyWriteCache()` to `beforeEach`, and added two new tests: one asserting 429 when the same username is written within the cooldown window, and one asserting 404 when the GitHub username does not exist.

## How to verify

1. Run `npx vitest run app/api/notify/route.test.ts` — all 23 tests pass (including the two new ones).
2. Send two rapid requests for the same username:
   ```
   curl -X POST http://localhost:3000/api/notify \
     -H "Content-Type: application/json" \
     -d '{"username": "octocat", "email": "a@b.com"}'
   ```
   First returns 200, second (within 5 minutes) returns 429 with a cooldown message.
3. Send a request for a non-existent GitHub username:
   ```
   curl -X POST http://localhost:3000/api/notify \
     -H "Content-Type: application/json" \
     -d '{"username": "thisuserdefinitelydoesnotexist123456", "email": "a@b.com"}'
   ```
   Returns 404 with "GitHub user not found."

## Edge cases considered

- **Cooldown scoped per-username, not per-IP** — An attacker rotating IPs cannot bypass the cooldown for a specific target user.
- **Cooldown only records on successful writes** — If the GitHub existence check fails or the database write fails, the cooldown is not set, so legitimate retries are not penalized.
- **No write amplification** — The cooldown is enforced before `dbConnect()` and before the MongoDB upsert, so a blocked request does not consume database resources.
- **DistributedCache fallback** — In environments without Redis/Vercel KV, the cooldown falls back to in-memory TTLCache, providing the same protection for single-instance deployments.
- **Out of scope** — Full GitHub OAuth authentication for the notify endpoint is not implemented here, as it requires a larger auth infrastructure change. The per-username cooldown + GitHub existence check mitigate the immediate authorization bypass.
- **Cooldown isolation from track-user** — This cooldown uses its own `DistributedCache` instance, separate from the cooldown in `track-user-protection.ts`, so the two endpoints do not interfere with each other.

Closes #3997

### DIFF
--- a/app/api/notify/route.test.ts
+++ b/app/api/notify/route.test.ts
@@ -16,9 +16,15 @@ vi.mock('@/lib/rate-limit', () => ({
     check: vi.fn().mockResolvedValue(true),
   },
 }));
+vi.mock('@/services/github/validate-user', () => ({
+  gitHubUserValidator: {
+    validateUser: vi.fn().mockResolvedValue(true),
+  },
+}));
 
 import { Notification } from '@/models/Notification';
 import { notifyRateLimiter } from '@/lib/rate-limit';
+import { gitHubUserValidator } from '@/services/github/validate-user';
 
 const makeRequest = (method: string, body?: object, search?: string) => {
   const url = `http://localhost:3000/api/notify${search ? '?' + search : ''}`;
@@ -36,6 +42,7 @@ describe('POST /api/notify', () => {
     vi.clearAllMocks();
     process.env = { ...originalEnv, MONGODB_URI: 'mongodb://localhost/test' };
     vi.mocked(notifyRateLimiter.check).mockResolvedValue(true);
+    vi.mocked(gitHubUserValidator.validateUser).mockResolvedValue(true);
   });
 
   afterEach(() => {
@@ -100,6 +107,38 @@ describe('POST /api/notify', () => {
     expect(res.status).toBe(429);
   });
 
+  // ── Per-username write cooldown ───────────────────────────────────────────
+
+  it('returns 429 when same username is written within cooldown period', async () => {
+    vi.mocked(Notification.findOneAndUpdate).mockResolvedValue({
+      username: 'cooldownuser',
+      email: 'a@b.com',
+      frequency: 'daily',
+      notifyOnCommit: true,
+      notifyOnStreak: true,
+      notifyOnMilestone: true,
+    } as never);
+
+    const body = { username: 'cooldownuser', email: 'a@b.com' };
+    const first = await POST(makeRequest('POST', body));
+    expect(first.status).toBe(200);
+
+    const second = await POST(makeRequest('POST', body));
+    expect(second.status).toBe(429);
+    const data = await second.json();
+    expect(data.message).toMatch(/Please wait \d+ seconds? before updating/);
+  });
+
+  // ── GitHub user existence check ──────────────────────────────────────────
+
+  it('returns 404 when GitHub username does not exist', async () => {
+    vi.mocked(gitHubUserValidator.validateUser).mockResolvedValue(false);
+    const res = await POST(makeRequest('POST', { username: 'nonexistent', email: 'a@b.com' }));
+    expect(res.status).toBe(404);
+    const data = await res.json();
+    expect(data.message).toContain('GitHub user not found');
+  });
+
   // ── MONGODB_URI handling ──────────────────────────────────────────────────
 
   it('returns 500 when MONGODB_URI is not set in production', async () => {
@@ -149,7 +188,7 @@ describe('POST /api/notify', () => {
 
   it('defaults frequency to daily and preferences to true when omitted', async () => {
     vi.mocked(Notification.findOneAndUpdate).mockResolvedValue({
-      username: 'testuser',
+      username: 'defaultuser',
       email: 'a@b.com',
       frequency: 'daily',
       notifyOnCommit: true,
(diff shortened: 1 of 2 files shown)


atul-upadhyay-7:
## Suggested labels

- **bug** — Authorization bypass vulnerability in POST /api/notify
- **security** — Unauthenticated endpoint accepted preferences for any GitHub username with no ownership verification
- **tests** — Added test coverage for per-username write cooldown and GitHub user existence check

Aamod007 (approved):

Issue #2439, opened 2026-06-01
Bug : Delta Sync Recalculates `totalContributions` and Discards GitHub's Authoritative Count


ArnavJoshi6391:
### Description

During delta synchronization, mergeCalendars() reconstructs totalContributions by summing the contribution counts of all merged days instead of preserving GitHub's authoritative contributionCalendar.totalContributions value.

Although GitHub already provides an authoritative totalContributions value in the GraphQL response, it is discarded and replaced with a client-side recomputation.

This introduces a data integrity risk because any discrepancy in the merged day set (missing dates, future merge logic changes, partial reconstruction issues, etc.) will cause displayed totals to diverge from GitHub's source-of-truth value.

The total contribution count should remain authoritative and come directly from GitHub rather than being recalculated locally during merge operations.

### Steps to Reproduce

1. Fetch contribution data for a user and cache the result.
2. Ensure delta sync is triggered (cached.calendar.lastSyncedAt exists and cache bypass is disabled).
3. Fetch a partial contribution window from GitHub.
4. Observe that mergeCalendars() rebuilds totalContributions using:
 sortedDays.reduce((sum, d) => sum + d.contributionCount, 0)
5. Notice that GitHub's GraphQL-provided contributionCalendar.totalContributions value is ignored during the merge process.

### Expected Behavior

totalContributions should preserve GitHub's authoritative value returned by the GraphQL API.

During delta sync, only day-level contribution data should be merged. The final calendar should retain the server-provided totalContributions count instead of recalculating it from merged day data.

### GitHub Username (If applicable)

ArnavJoshi6391

### Environment

Chrome

PR #4099, opened 2026-06-05
fix(github): replace partial delta sync with a full-window refresh

### LINKED ISSUE #3555: bug: delta-sync uses partial-window data for totals and overwrites repoContributions (regression from #2447)
## Description

After the 5-minute cache freshness window, `/api/streak` (and `/api/stats`, the dashboard, and GitHub Wrapped) can display stats built from a partial refresh window instead of the full year. The shared root cause is the delta-sync logic in `fetchContributionsUncached` (`lib/github.ts`), which re-queries only `[lastSyncedAt - 1 day, to]` and then treats that narrow window's data as if it were the full window. There are three concrete symptoms.

### 1. `totalContributions` collapses (regression from #2447)

During delta sync, the GraphQL `contributionCalendar.totalContributions` returned for the narrow window is the partial-window total (for example the last day's contributions), not the full year. PR #2447 (fixing #2439) changed `mergeCalendars` to preserve GitHub's `contributionCalendar.totalContributions` rather than recomputing it from the merged days. That is correct for a full fetch, but during delta sync the value passed in is the partial-window total, so the merged calendar's `totalContributions` is set to that partial value. The day-level `weeks` data is merged correctly (full history is preserved), but the headline total collapses.

### 2. PR and Issue totals drift upward (double-counting)

In the same delta block, `totalPRs` and `totalIssues` are taken from the narrow-window query and then the cached values are added on top:

```
totalPRs += cached.totalPRs || 0;
totalIssues += cached.totalIssues || 0;
```

The merged result is re-cached under the same key, so every delta refresh (every 5+ minutes) adds the latest window on top of the already-accumulated total, and the deliberately-overlapping boundary day (`lastSyncedAt - 1 day`) is counted again. `stats.totalPRs` / `stats.totalIssues` therefore grow without bound rather than reflecting GitHub's actual totals.

### 3. `repoContributions` is overwritten with delta-only data

`commitContributionsByRepository` from the narrow-window query covers only the delta window, and it is cached as-is (it is not merged with the cached full set). After a delta refresh the dashboard's repository contribution data and language breakdown collapse to whatever happened in the last day or two.

### Secondary issue: `from > to` for historical requests

For historical / fixed-window requests (`?year=`, or `from` + `to`), delta sync still overrides `queryFrom` to `lastSyncedAt - 1 day` while keeping the original `to`. When `lastSyncedAt - 1 day` is after `to` (for example when requesting a past year), the query becomes `from > to` and is invalid.

### Fix is a maintainer design decision (interacts with merged #2447 / closed #2439)

- Reconcile the delta values against the cached authoritative data (for example total = `cachedTotal - overlappingOldDays + deltaWindowTotal`; assign full-window aggregate PR/issue totals rather than `+=`; merge repoContributions), keeping delta sync but computing correct full-window values.
- Or stop issuing the narrow delta query for the rolling window and do a full refetch on refresh, so GitHub's returned `totalContributions`, PR/issue totals, and repoContributions are already the full authoritative values. This also removes the `from > to` break and the unbounded growth, at the cost of the delta-sync optimization.

## Steps to Reproduce

1. Run the app with a valid `GITHUB_TOKEN`, single instance (in-memory cache; with Redis the same applies cluster-wide).
2. `GET /api/streak?user=octocat&format=json` (or load `/dashboard/octocat`) and note `totalContributions`, `totalPRs`, and `totalIssues`. The entry is cached with `lastSyncedAt = now`.
3. Wait more than 5 minutes (the freshness threshold) but less than 7 days, then request again on the same warm instance. The freshness check forces a delta refresh.
4. Observe: `totalContributions` collapses toward the last day or two; `totalPRs` / `totalIssues` exceed the true totals and grow on each further refresh; the contributed-repos / language breakdown reflects only the delta window.

For the historical case: `GET /api/streak?user=octocat&year=2023&format=json` after a recent sync can produce a `from > to` delta query.

## Expected Behavior

`totalContributions`, `totalPRs`, `totalIssues`, and `repoContributions` continue to reflect the full requested window (the annual total for the default view, the selected year for `?year=`), regardless of cache freshness or delta sync. Aggregate totals are not accumulated across refreshes, and fixed-window / historical requests are never turned into `from > to` queries.

## Screenshots / Logs

Not applicable. Server-side caching behavior.

## Environment

Other

Anexus5919:
## Description

Fixes #3555

After the 5-minute freshness window, `fetchContributionsUncached` (`lib/github.ts`) issued a "delta sync": it re-queried only `[lastSyncedAt - 1 day, to]` and then treated that narrow window as if it were the full window. That produced four defects:

1. `totalContributions` collapsed: `mergeCalendars` was given the narrow window's `totalContributions` as the authoritative value (regression from #2447), so the headline total dropped to the partial-window figure.
2. `totalPRs` / `totalIssues` drifted upward: the narrow-window totals were added on top of the cached totals (`+=`) on every refresh, double-counting the overlapping boundary day and growing without bound.
3. `repoContributions` was overwritten with delta-only data, collapsing the repository and language breakdown to the last day or two.
4. For historical requests (`?year=`, or `from` + `to`), `queryFrom` was overridden to `lastSyncedAt - 1 day` while keeping the original `to`, which could produce an invalid `from > to` query.

These aggregates are window-scoped and cannot be reconciled from delta data alone (the boundary day overlaps and there is no per-day PR/issue/repo breakdown to subtract), so the correct source is a full-window query. This change removes the delta-sync path: every refresh now re-queries the full requested window (`from = options.from`, never rewritten), so `totalContributions`, `totalPRs`, `totalIssues`, and `repoContributions` come straight from GitHub's authoritative full-window response. `mergeCalendars` is removed as it is no longer used.

This also satisfies #2439 (the headline total is GitHub's authoritative value and is never recomputed from merged days), and removes the `from > to` break. `lastSyncedAt` is still stamped on every fetch and still drives the 5-minute freshness check and the background-refresh staleness check; the 7-day cache still backs the offline/stale fallback. A GraphQL `contributionsCollection` query costs the same regardless of window, so there is no rate-limit cost to querying the full window.

Behavior note: the cached calendar is now exactly the window GitHub returns for each request and no longer accumulates days across refreshes, so days that roll out of GitHub's trailing window are dropped on refresh rather than retained. The default streak, stats, OG, and dashboard views only ever display the trailing window, so this is the intended result of the fix.

Added tests in `lib/github.test.ts` that drive a refresh after expiry and assert the full window is re-queried with authoritative totals (no collapse, no accumulation, no partial repo overwrite), and that a historical fixed-window refresh re-queries the requested window and never produces `from > to`. Both tests were verified to fail against the previous delta-sync code.

## Pillar

## Visual Preview

Not applicable. Backend data-fetch and caching logic change with no UI output.

- [x] I have tested these changes locally (full Vitest suite passes; `lib/github.test.ts` 144 tests pass including the new ones; `tsc --noEmit` clean).
- [ ] The SVG output matches the CommitPulse "premium quality" aesthetic standard. (Not applicable, no SVG output in this change.)

### DIFF
--- a/lib/github.test.ts
+++ b/lib/github.test.ts
@@ -1633,15 +1633,82 @@ describe('GitHub API cache behavior', () => {
     expect(fetch).toHaveBeenCalledTimes(2);
   });
 
-  it('cache expiry: expired entry triggers a delta sync fetch', async () => {
+  it('cache expiry: refresh re-queries the full window with authoritative totals, no collapse or accumulation or partial overwrite', async () => {
     vi.useFakeTimers();
     vi.setSystemTime(new Date('2026-01-01T00:00:00.000Z'));
+
+    const buildResponse = (total: number, prs: number, issues: number, languages: string[]) =>
+      mockResponse({
+        data: {
+          user: {
+            contributionsCollection: {
+              totalPullRequestContributions: prs,
+              totalIssueContributions: issues,
+              contributionCalendar: {
+                totalContributions: total,
+                weeks: [
+                  {
+                    contributionDays: [
+                      { contributionCount: total, date: '2025-12-31', color: '#216e39' },
+                    ],
+                  },
+                ],
+              },
+              commitContributionsByRepository: languages.map((name) => ({
+                repository: { primaryLanguage: { name } },
+                contributions: { totalCount: total },
+              })),
+            },
+          },
+        },
+      });
+
+    // Full window (from undefined) returns authoritative annual data; a narrow window returns partial data.
+    vi.mocked(fetch).mockImplementation(async (_url, init) => {
+      const body = JSON.parse((init as RequestInit).body as string);
+      return body.variables.from === undefined
+        ? buildResponse(120, 12, 6, ['TypeScript', 'Go', 'Rust', 'Python'])
+        : buildResponse(5, 1, 2, ['TypeScript']);
+    });
+
+    const first = await fetchGitHubContributions('octocat');
+    expect(first.calendar.totalContributions).toBe(120);
+    expect(first.totalPRs).toBe(12);
+    expect(first.totalIssues).toBe(6);
+    expect(first.repoContributions).toHaveLength(4);
+
+    vi.setSystemTime(Date.now() + GITHUB_CACHE_TTL_MS + 1);
+    const refreshed = await fetchGitHubContributions('octocat');
+
+    expect(fetch).toHaveBeenCalledTimes(2);
+
+    // The refresh must query the full window, never a narrow delta window.
+    const secondCallBody = JSON.parse(vi.mocked(fetch).mock.calls[1][1]!.body as string);
+    expect(secondCallBody.variables.from).toBeUndefined();
+
+    // Full-window values survive the refresh: no collapse, no accumulation, no delta-only repo overwrite.
+    expect(refreshed.calendar.totalContributions).toBe(120);
+    expect(refreshed.totalPRs).toBe(12);
+    expect(refreshed.totalIssues).toBe(6);
+    expect(
+      refreshed.repoContributions.map((r) => r.repository.primaryLanguage?.name).sort()
+    ).toEqual(['Go', 'Python', 'Rust', 'TypeScript']);
+  });
+
+  it('historical fixed-window refresh re-queries the requested window, never from > to', async () => {
+    vi.useFakeTimers();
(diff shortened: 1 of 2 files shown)


Anexus5919:
@Aamod007 @JhaSourav07 Kindly have a review on this pr. Thanks!

Aamod007 (approved):

Aamod007 (changes requested):
.

Aamod007 (approved):

PR #4709, opened 2026-06-06
test(cache): verify DistributedCache rejects negative TTL before Redis write

### LINKED ISSUE #1388: test(cache): verify TTLCache behavior for negative TTL (Variation 1)
## 🧩 Background
The existing cache implementation must handle boundary inputs robustly. Currently, tests do not comprehensively cover extreme values.

## 🎯 Objective
Add a specific test to verify cache behavior for negative TTL.

## 📁 Files to touch
- `lib/cache.test.ts`

## 🛠️ Implementation steps
- Assert that setting a value with -5000 throws a RangeError.
- Verify verify that setting a value with a negative TTL is rejected.
- Run Vitest to ensure the test coverage is successfully verified.

## ✅ Definition of done
- [ ] 1 new test case targeting this boundary in cache.
- [ ] All cache test suites pass successfully.

ARPANPATRA111:
## Description

Closes #1388

Adds one focused test for the `-5000` TTL boundary on the **production code path** — `DistributedCache.set`, the cache actually used by `/api/streak` and the GitHub contribution fetch.

### Why this is not a redundant test

The synchronous `TTLCache.set` rejection for negative TTL is already covered (`throws RangeError when setting a value with -5000 TTL` and `throws RangeError when ttlMs is 0 or negative`). Re-asserting the same guard adds no coverage.

This test instead pins down runtime behaviour of `DistributedCache` that nothing else covers:

1. A `-5000` TTL is rejected with a `RangeError` (`ttlMs must be positive, got -5000`).
2. **No Redis REST call is issued** when the TTL is invalid — `fetch` is never called. This is the meaningful guarantee: without the early guard an invalid TTL could orphan an entry in the shared Redis store while the local L1 cache stays empty, silently desynchronising the two layers.
3. The instance remains usable — a subsequent valid `set` resolves and issues exactly one Redis `SET`.

Removing the validation in `DistributedCache.set` makes this test fail, so it locks in real behaviour rather than restating an existing assertion.

## Checklist

- [x] 1 new test targeting the `-5000` boundary on the production `DistributedCache` path
- [x] All cache test suites pass (`npx vitest run lib/cache.test.ts` → 63 passed)
- [x] Prettier clean
- [x] ESLint clean
- [x] No production code changed — tests only


Aamod007 (changes requested):
### Summary

The PR follows a known pattern of AI-generated superficial test files or redundant 'complementary tests' that provide little to no meaningful runtime coverage and clutter the repository.

### Required Changes

#### Issue 1

* Problem: The test file creates or simulates interactions on dummy structures, or simply adds redundant 'complementary tests' without real feature value.
* Impact: It provides 0 real test coverage or adds maintenance overhead without fixing actual application bugs.
* Required Fix: Ensure the PR tests actual new runtime behavior and does not just add redundant tests.

ARPANPATRA111 (commented):
@Aamod007 review it once 👍

PR #4710, opened 2026-06-06
test(tz): reject path-traversal and injection ?tz= payloads at schema and route

### LINKED ISSUE #1449: test(api validation): check query validation boundaries for ?tz= parameter (Variation 3)
## 🧩 Background
Input validation prevents arbitrary code and path traversal sequences from reaching the application core logic.

## 🎯 Objective
Write a schema-level validation test targeting the `?tz=` parameter under valid IANA timezone check.

## 📁 Files to touch
- `app/api/streak/route.test.ts`
- `lib/validations.test.ts`

## 🛠️ Implementation steps
- Mock a request passing the invalid value: `'Mars/Cyonia'`.
- Assert that the validation handler returns a 400 Bad Request with 'Invalid timezone'.

## ✅ Definition of done
- [ ] 1 new test case targeting the query input parameter.
- [ ] All endpoint test suites pass successfully.

ARPANPATRA111:
## Description

Closes #1449

The plain `?tz=Mars/Cyonia` rejection is already covered at both layers — the route (`route.test.ts` "fictitious planetary timezone") and the schema (`validations.test.ts` "rejects invalid IANA timezone names", plus a second variation). Re-asserting it adds no runtime coverage.

This PR instead targets the threat the issue's background actually names — **path-traversal sequences and arbitrary code** reaching the application core logic — which no existing `?tz=` test exercises.

### `lib/validations.test.ts` (schema-level, the file the issue names)
`streakParamsSchema.safeParse` rejects a path-traversal payload (`../../../../etc/passwd`), a command-injection payload (`America/New_York; rm -rf /`) and a whitespace-padded zone, each with the exact `Invalid timezone` message — while a real IANA zone (`America/New_York`) still parses. The valid-zone assertion proves the `isValidTimeZone` refine *discriminates* rather than blanket-failing.

### `app/api/streak/route.test.ts` (end-to-end runtime)
A path-traversal `?tz=` returns a `400` `Invalid parameters` envelope **and never reaches the side-effecting core logic** — neither `fetchGitHubContributions` (the GitHub API) nor `getSecondsUntilMidnightInTimezone` (the cache-TTL helper) is called. No prior `?tz=` test asserts this guard, so it locks in that a malformed timezone cannot spend a rate-limit token or flow into TTL computation.

Removing the `isValidTimeZone` refine makes both tests fail, so they pin real behavior rather than restating the existing Mars/Cyonia assertion.

## Checklist

- [x] New test cases targeting the `?tz=` query parameter (schema + route)
- [x] All endpoint test suites pass (`npx vitest run app/api/streak/route.test.ts lib/validations.test.ts` → 364 passed)
- [x] Prettier clean
- [x] ESLint clean
- [x] No production code changed — tests only


Aamod007 (changes requested):
### Summary

The PR follows a known pattern of AI-generated superficial test files or redundant 'complementary tests' that provide little to no meaningful runtime coverage and clutter the repository.

### Required Changes

#### Issue 1

* Problem: The test file creates or simulates interactions on dummy structures, or simply adds redundant 'complementary tests' without real feature value.
* Impact: It provides 0 real test coverage or adds maintenance overhead without fixing actual application bugs.
* Required Fix: Ensure the PR tests actual new runtime behavior and does not just add redundant tests.

PR #4777, opened 2026-06-07
fix(streak): trust the sanitized speed so in-range decimals are preserved

### LINKED ISSUE #4775: bug: Streak route re-validates the already-sanitized speed with an integer-only regex, dropping valid decimals
## Description

The `/api/streak` route validates the `speed` query parameter twice. The zod schema (`streakParamsSchema` in `lib/validations.ts`) already sanitizes it via `sanitizeSpeed`, which accepts any value matching `\d+(\.\d+)?s` in the range 2s to 20s, so a request like `?speed=8.5s` yields the valid sanitized value `8.5s`.

When building `BadgeParams`, the route then re-validates that same value with an integer-only regex `/^(?:[2-9]|1\d|20)s$/`, which does not match decimals. A valid in-range decimal such as `8.5s` therefore fails the second check and falls back to `8s`. Valid decimal speeds accepted by the schema are silently dropped.

## Steps to Reproduce

1. Request `/api/streak?user=octocat&speed=8.5s`.
2. Inspect the returned SVG (the `--scan-speed` animation duration).
3. Observe it is `8s` rather than the requested `8.5s`, even though `8.5s` is within the accepted 2s to 20s range.

## Expected Behavior

The route trusts the value already sanitized by the schema, so a valid in-range decimal speed like `8.5s` is preserved instead of being reset to `8s`.

## Screenshots / Logs

Not applicable. Server-side parameter handling.

## Environment

Other

Anexus5919:
## Description

Fixes #4775

The `/api/streak` route validated the `speed` parameter twice. `streakParamsSchema` already sanitizes it through `sanitizeSpeed`, which accepts `\d+(\.\d+)?s` in the 2s to 20s range, so `?speed=8.5s` produces the valid value `8.5s`. The route then re-checked that value while building `BadgeParams` with an integer-only regex `/^(?:[2-9]|1\d|20)s$/`, which does not match decimals, so a valid `8.5s` was discarded and reset to `8s`.

This change trusts the value already sanitized by the schema and assigns `speed` directly. `SpeedString` is `` `${number}s` `` and `sanitizeSpeed` guarantees a value in range (or the `8s` default), so the result is still a safe CSS duration with no second validation needed.

Updated the test that asserted the old behavior (a decimal was forced to `8s`) to instead assert that an in-range decimal like `8.5s` is preserved. The existing fallback tests (invalid format, missing unit, below or above bounds, and decimals below the minimum) are unchanged and still pass, because `sanitizeSpeed` already rejects those.

## Pillar

## Visual Preview

No new UI. A badge requested with an in-range decimal speed (for example `?speed=8.5s`) now animates at that speed instead of being reset to `8s`.

- [x] I have tested these changes locally (all `app/api/streak/route.test.ts` tests pass, 182 tests; `tsc --noEmit` clean).
- [ ] The SVG output matches the CommitPulse "premium quality" aesthetic standard. (Server-side parameter fix, no SVG layout change.)

### DIFF
--- a/app/api/streak/route.test.ts
+++ b/app/api/streak/route.test.ts
@@ -481,11 +481,10 @@ describe('GET /api/streak', () => {
       expect(body).toContain('20s');
     });
 
-    it('falls back to 8s when speed is a non-integer decimal like "2.0s"', async () => {
-      const response = await GET(makeRequest({ user: 'octocat', speed: '2.0s' }));
+    it('preserves an in-range decimal speed like "8.5s" instead of forcing it to an integer', async () => {
+      const response = await GET(makeRequest({ user: 'octocat', speed: '8.5s' }));
       const body = await response.text();
-      expect(body).toContain('8s');
-      expect(body).not.toContain('2.0s');
+      expect(body).toContain('8.5s');
     });
   });
 
--- a/app/api/streak/route.ts
+++ b/app/api/streak/route.ts
@@ -202,7 +202,7 @@ export async function GET(request: Request) {
       accent: isAutoTheme ? selectedTheme.accent : accent || selectedTheme.accent,
       border: sanitizedBorder,
       radius,
-      speed: speed && /^(?:[2-9]|1\d|20)s$/.test(speed) ? speed : '8s',
+      speed,
       scale,
       font,
       autoTheme: isAutoTheme,


Anexus5919:
@Aamod007 @JhaSourav07 Kindly have a review on this pr. Thanks!

Aamod007 (approved):

PR #4835, opened 2026-06-07
fix(wrapped): trust the sanitized speed so in-range decimals are preserved

### LINKED ISSUE #4833: bug: Wrapped route re-validates the already-sanitized speed with an integer-only regex, dropping valid decimals
## Description

The `/api/wrapped` route validates the `speed` query parameter twice. The zod schema (`wrappedParamsSchema` in `lib/validations.ts`) already sanitizes it via `sanitizeSpeed`, which accepts any value matching `\d+(\.\d+)?s` in the range 2s to 20s, so a request like `?speed=8.5s` yields the valid sanitized value `8.5s`.

When building `BadgeParams`, the route then re-validates that same value with an integer-only regex `/^(?:[2-9]|1\d|20)s$/`, which does not match decimals. A valid in-range decimal such as `8.5s` therefore fails the second check and falls back to `8s`. Valid decimal speeds accepted by the schema are silently dropped.

This is the same defect that was fixed for `/api/streak`; it also exists in the Wrapped route.

## Steps to Reproduce

1. Request `/api/wrapped?user=octocat&speed=8.5s`.
2. Inspect the returned SVG (the `--scan-speed` animation duration).
3. Observe it is `8s` rather than the requested `8.5s`, even though `8.5s` is within the accepted 2s to 20s range.

## Expected Behavior

The route trusts the value already sanitized by the schema, so a valid in-range decimal speed like `8.5s` is preserved instead of being reset to `8s`.

## Screenshots / Logs

Not applicable. Server-side parameter handling.

## Environment

Other

Anexus5919:
## Description

Fixes #4833

The `/api/wrapped` route validated the `speed` parameter twice. `wrappedParamsSchema` already sanitizes it through `sanitizeSpeed`, which accepts `\d+(\.\d+)?s` in the 2s to 20s range, so `?speed=8.5s` produces the valid value `8.5s`. The route then re-checked that value while building `BadgeParams` with an integer-only regex `/^(?:[2-9]|1\d|20)s$/`, which does not match decimals, so a valid `8.5s` was discarded and reset to `8s`.

This is the same defect already fixed for `/api/streak`. This change applies the same fix to the Wrapped route: trust the value already sanitized by the schema and assign `speed` directly. `SpeedString` is `` `${number}s` `` and `sanitizeSpeed` guarantees a value in range (or the `8s` default), so the result is still a safe CSS duration with no second validation needed.

Added speed tests to `app/api/wrapped/route.test.ts`: an in-range decimal like `8.5s` is preserved, and out-of-range or malformed values still fall back to `8s` (handled by `sanitizeSpeed`).

## Pillar

## Visual Preview

No new UI. A Wrapped image requested with an in-range decimal speed (for example `?speed=8.5s`) now animates at that speed instead of being reset to `8s`.

- [x] I have tested these changes locally (all `app/api/wrapped/route.test.ts` tests pass, 22 tests including 2 new speed tests; `tsc --noEmit` clean).
- [ ] The SVG output matches the CommitPulse "premium quality" aesthetic standard. (Server-side parameter fix, no SVG layout change.)


Anexus5919:
@JhaSourav07 @Aamod007 Kindly have a review on this pr. Thanks!

Anexus5919:
Heads-up on the failing Format / Lint / Typecheck / Test check: it is not caused by this PR. The change here is limited to `app/api/wrapped/route.ts` and its test, and the wrapped route suite passes locally (22 tests, including the new `8.5s` decimal and fallback cases). The check is red because of two pre-existing failures on `main`:

1. `components/dashboard/HistoricalTrendView.empty-fallback.test.tsx` asserts a fixed `Current Streak` label, but that label is now date-relative (`Current` / `Ending` / `Upcoming` depending on where the selected window sits versus today), so it fails once the test's January 2026 window is in the past. I have opened issue #4837 and PR #4838 to make that assertion label-agnostic.

2. The Typecheck step fails with `TS2307: Cannot find module './proxy'` from `proxy.test.ts` and `proxy.accessibility.test.ts`, which import a `./proxy` module that is not present in the repo. This one is separate from this PR and from #4838; it needs `proxy.ts` to be restored (or those orphaned tests removed) on `main`.

Once both are resolved on `main`, a rebase here should bring the check green.

Aamod007 (approved):
﻿Excellent logic fix! Preserving in-range decimals while trusting the sanitized speed ensures the Wrapped statistics remain precise. Approving!

PR #4938, opened 2026-06-08
[Bug] Race condition in DistributedCache.getOrSet — concurrent calls both execute loadFn

### LINKED ISSUE #4928: [Bug] Race condition in DistributedCache.getOrSet — concurrent calls both execute loadFn despite local locks
## Description

The `DistributedCache.getOrSet` method in `lib/cache.ts` has a race condition — concurrent calls for the same cache key both execute the `loadFn` instead of one waiting for the other, defeating the purpose of local lock deduplication.

## Cause

At line 468, `await this.get(key)` yields the event loop. Between this await completing and the local lock check/set at lines 476 and 568, a second concurrent call can interleave:

1. Call A: awaits `this.get(key)` → yields event loop
2. Call B: awaits `this.get(key)` → yields event loop
3. Call A: resumes, `localLocks.get(key)` → `undefined`
4. Call B: resumes, `localLocks.get(key)` → `undefined` (A hasn't set it yet)
5. Call A: sets `localLocks.set(key, promise)` at line 568
6. Call B: sets `localLocks.set(key, promise)` at line 568 → overwrites A's promise
7. Both A and B execute `loadFn` independently

## Fix

Move the local lock check to happen synchronously before any `await`, so the second caller always finds and returns the existing in-flight promise. See the PR for full implementation.

## Additional Context

In serverless environments with many concurrent instances, a cache expiry for a popular user triggers N simultaneous fetches instead of 1, multiplying API/database load. This affects all callers of `getOrSet` across the codebase.

## Follow-up

This is a correctness + performance fix. No breaking API changes.

AnushKamble:
## Issue
#4928 — Race condition in \DistributedCache.getOrSet\ where concurrent calls for the same cache key both execute \loadFn\ instead of sharing the result.

## Cause
\wait this.get(key)\ at line 468 yields the event loop. Between this await completing and \	his.localLocks.set(key, promise)\ at line 568, a second concurrent call can interleave and also miss the local lock check, resulting in both calls executing \loadFn\ independently.

## Fix
1. **Pre-await check**: Check \	his.localLocks.get(key)\ synchronously before the first \wait\, so a call that arrives while another is mid-flight returns the existing promise immediately.
2. **Post-await double-check**: Re-check \	his.localLocks\ after the \wait this.get(key)\, catching the window where a second call entered before the promise was stored.

Since all operations between a resumed await and the next yield are synchronous (block the event loop), the promise creation and lock storage are atomic from the perspective of other callers.

## Additional Context
In serverless environments with concurrent instances, a cache expiry for a popular user triggers N simultaneous fetches instead of 1, multiplying API/database load.

Closes #4928

### DIFF
--- a/lib/cache.ts
+++ b/lib/cache.ts
@@ -467,6 +467,11 @@ return c`;
     ttlMs: number,
     shouldFetch?: (cached: T) => boolean
   ): Promise<T> {
+    // Join an existing in-flight request before any async operation to avoid
+    // concurrent loadFn execution for the same key.
+    const existing = this.localLocks.get(key);
+    if (existing) return existing;
+
     // Attempt to retrieve an existing value before triggering a refresh.
     const cached = await this.get(key);
 
@@ -474,8 +479,7 @@ return c`;
       return cached;
     }
 
-    // Join an existing in-flight request instead of creating duplicate fetches
-    // within the same runtime instance.
+    // Double-check local locks after the await in case another call interleaved.
     const pendingLocal = this.localLocks.get(key);
     if (pendingLocal) return pendingLocal;
 


AnushKamble:
@JhaSourav07 Branch is rebased against latest upstream/main and ready for merging.

Aamod007 (approved):
﻿Awesome debugging! Resolving the race condition in DistributedCache.getOrSet prevents duplicate expensive data fetches from executing concurrently. Approving!

PR #5126, opened 2026-06-10
fix: Race Condition in DistributedCache.update() Affecting Rate Limiting Consistency

### LINKED ISSUE #4761: bug: Race condition in DistributedCache.update() causes inconsistent rate limiting across instances
### Description

A race condition exists between DistributedCache and RateLimiter when cache entries expire between get() and update() operations.

DistributedCache.update() updates both the local cache (L1) and the distributed store (Redis/KV). In distributed mode it relies on Redis SET ... KEEPTTL XX semantics, which fail if the key has already expired.

RateLimiter.checkWithResult() performs a get() followed by update(), but it does not check the boolean result returned by update(). If the key expires between these operations, the update may fail silently, causing the rate-limit state to become inconsistent across instances.

This issue is most likely to appear in production environments with multiple serverless instances, network latency, or requests occurring near TTL expiration boundaries.

### Steps to Reproduce

Configure the application to use Redis/KV as the distributed cache backend.
Deploy multiple application instances.
Create a rate-limit record with a short TTL.
Send requests near the expiration boundary.
Allow the key to expire between cache.get() and cache.update().
Observe that update() returns false, but the result is ignored by RateLimiter.checkWithResult().
Send additional requests from different instances and observe inconsistent rate-limit behavior.

### Expected Behavior

If an update fails because the key expired, the rate limiter should either:

Retry or recreate the record,
Use an atomic increment operation,
Or explicitly handle the update failure.

Requests should not be allowed or denied based on stale or partially updated cache state.

### Screenshots / Logs

_No response_

_No response_

### Environment

Chrome

Krishnx21:
## Related Issue

Closes #4761 

## Description

This PR fixes a race condition between `DistributedCache` and `RateLimiter` that can cause inconsistent rate-limiting behavior across distributed application instances.

### Root Cause

`RateLimiter.checkWithResult()` currently performs the following sequence:

```text
cache.get()
   ↓
modify value
   ↓
cache.update()
```

In distributed cache mode, `DistributedCache.update()` relies on Redis/KV update semantics (`SET ... KEEPTTL XX`), which only succeed if the key still exists.

If the cache entry expires between the `get()` and `update()` operations:

1. `get()` returns a valid rate-limit record.
2. The key expires before `update()` executes.
3. `update()` returns `false`.
4. The return value is ignored by `RateLimiter.checkWithResult()`.
5. Rate-limit state becomes inconsistent across instances.

This can result in requests being incorrectly allowed or denied when multiple instances are processing traffic simultaneously.

### Rate Limiter Improvements

* Added validation for the return value of `DistributedCache.update()`.
* Prevented silent failures when cache updates do not succeed.
* Added handling for expired records detected during update operations.

### Cache Consistency Improvements

* Ensured update failures are explicitly handled.
* Recreated or refreshed rate-limit records when necessary.
* Improved synchronization between local cache (L1) and distributed cache (Redis/KV).

### Error Handling

* Added safeguards for TTL-boundary edge cases.
* Prevented stale state propagation across distributed instances.

### Before

```text
Instance A:
get() → success

Key expires

update() → false

Result ignored

Rate limit state becomes inconsistent
```

### After

```text
Instance A:
get() → success

Key expires

update() → false

Failure detected

Record recreated/retried

State remains consistent
```

### Reproduction Scenario

1. Configure Redis/KV as the distributed cache backend.
2. Deploy multiple application instances.
3. Create a rate-limit entry with a short TTL.
4. Generate requests close to expiration boundaries.
5. Force expiration between `get()` and `update()`.
6. Verify that update failures are handled correctly.

### Results

* Update failures no longer occur silently.
* Rate-limit records remain consistent across instances.
* No incorrect allow/deny decisions observed near TTL expiration boundaries.
* Distributed deployments behave predictably under concurrent load.

## Impact

**Severity:** Medium–High

This fix improves:

* Distributed cache consistency
* Rate-limiting accuracy
* Multi-instance reliability
* Production stability under concurrent traffic
* Resilience to TTL expiration edge cases

## Checklist

* [x] Issue reproduced
* [x] Root cause identified
* [x] Update failure handling implemented
* [x] Distributed cache behavior validated
* [x] Concurrent scenarios tested

---

## GSSoC 2026

**Contributor:** @Krishnx21

**Program:** GirlScript Summer of Code 2026 (GSSoC'26)

### DIFF
--- a/lib/cache.test.ts
+++ b/lib/cache.test.ts
@@ -868,6 +868,37 @@ describe('DistributedCache', () => {
     );
     cache.destroy();
   });
+
+  it('evicts stale local state when a Redis update loses an expiry race', async () => {
+    process.env.KV_REST_API_URL = 'https://mock-redis.upstash.io';
+    process.env.KV_REST_API_TOKEN = 'mock-token';
+
+    vi.mocked(fetch)
+      .mockResolvedValueOnce({
+        ok: true,
+        status: 200,
+        json: () => Promise.resolve({ result: 'OK' }),
+      } as Response)
+      .mockResolvedValueOnce({
+        ok: true,
+        status: 200,
+        json: () => Promise.resolve({ result: null }),
+      } as Response)
+      .mockResolvedValueOnce({
+        ok: true,
+        status: 200,
+        json: () => Promise.resolve({ result: null }),
+      } as Response);
+
+    const cache = new DistributedCache<string>();
+    await cache.set('redis-key', 'old-value', 60000);
+
+    expect(await cache.update('redis-key', 'new-value')).toBe(false);
+    expect(await cache.get('redis-key')).toBeNull();
+    expect(fetch).toHaveBeenCalledTimes(3);
+
+    cache.destroy();
+  });
 });
 
 describe('TTLCache with infinite TTL', () => {
--- a/lib/cache.ts
+++ b/lib/cache.ts
@@ -350,10 +350,8 @@ export class DistributedCache<T> {
   }
 
   async update(key: string, value: T): Promise<boolean> {
-    this.localCache.update(key, value);
-
     if (!this.useRedis) {
-      return this.localCache.has(key);
+      return this.localCache.update(key, value);
     }
 
     try {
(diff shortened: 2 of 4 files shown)


Aamod007 (approved):
﻿Excellent fix! Recreating the rate-limit record when it expires mid-operation prevents silent failures and keeps distributed cache deployments consistent. Very thorough catch. Approving!

Krishnx21:
Thank you for reviewing and approving the changes!

PR #5206, opened 2026-06-11
fix: prevent rate-limit bypass via spoofed IP headers

### LINKED ISSUE #5196: bug: Rate-Limit Bypass via Spoofed Client IP Headers
### Description

The application's rate-limiting mechanism relies on client-controlled IP headers, including x-real-ip and cf-connecting-ip, without verifying that the request originated from a trusted proxy.

Because these headers are fully controllable by the client, an attacker can continuously rotate header values and obtain a new rate-limit bucket for every request, effectively bypassing rate limiting.

Additionally, the IP resolution logic can resolve certain requests to 127.0.0.1. The track-user endpoint explicitly skips rate limiting for localhost, allowing a complete bypass of protection mechanisms.

This issue can be abused to generate excessive API traffic, consume GitHub API quotas, trigger expensive processing operations, and increase the risk of service degradation.

Affected Locations:

getClientIp.ts (line 147)
IP resolution logic (line 90)
track-user route (line 14)

### Steps to Reproduce

 Header Rotation
Send requests to a rate-limited endpoint.
Include a custom x-real-ip or cf-connecting-ip header.
Change the header value for each request.
Observe that requests continue to succeed despite expected rate limits.

### Expected Behavior

Rate limiting should be based on a trusted client IP source.

Forwarding headers should only be honored when requests originate from explicitly configured reverse proxies.

Untrusted clients should not be able to influence rate-limit bucket assignment.

Localhost exemptions should not be reachable through user-controlled headers.

### Screenshots / Logs

_No response_

_No response_

### Environment

Chrome

ASSIGN ME UNDER GSSOC 2026

Krishnx21:
## Summary

- remove middleware's direct trust of the client-controlled `x-real-ip` header
- treat forwarding headers as untrusted unless a separately supplied direct peer IP is a configured trusted proxy
- use authoritative `NextRequest.ip` when the platform provides it
- return the direct peer IP when it is not trusted instead of allowing forwarded-header overrides
- remove the `127.0.0.1` rate-limit exemption from `/api/track-user`
- add regression tests for rotating `x-real-ip`, spoofed localhost headers, authoritative request IPs, and localhost rate limiting

## Security impact

Previously, attackers could rotate `x-real-ip` or other forwarding headers to receive a fresh rate-limit bucket for every request. Some spoofed requests could also resolve to `127.0.0.1`, causing `/api/track-user` to skip its limiter entirely.

Forwarding headers can no longer establish their own trust boundary, and localhost requests now consume rate-limit capacity like every other request.

## Verification

- `npm run test -- middleware app/api/track-user utils services/security lib/rate-limit.test.ts` (42 files, 333 tests passed)
- `npm run test -- utils/getClientIp.test.ts utils/trustedProxy.test.ts middleware.test.ts middleware.accessibility.test.ts app/api/track-user/route.test.ts lib/rate-limit.test.ts` (6 files, 84 tests passed)
- `npm run typecheck`
- focused ESLint and `git diff --check`

`npm run build` could not start compilation because Windows/OneDrive denied Next.js permission to remove the pre-existing generated `.next/diagnostics` directory (`EPERM`).

Fixes #5196

### DIFF
--- a/app/api/track-user/route.test.ts
+++ b/app/api/track-user/route.test.ts
@@ -176,6 +176,12 @@ describe('POST /api/track-user', () => {
     expect(response.headers.get('x-ratelimit-reset')).toBe(reset.toString());
   });
 
+  it('applies rate limiting to localhost requests', async () => {
+    await POST(makeRequest({ username: 'valid-user' }));
+
+    expect(trackUserRateLimiter.checkWithResult).toHaveBeenCalledWith('127.0.0.1');
+  });
+
   describe('Without MONGODB_URI (Local Development Bypass)', () => {
     it('returns 200 and bypassed flag when MONGODB_URI is undefined', async () => {
       delete process.env.MONGODB_URI;
--- a/app/api/track-user/route.ts
+++ b/app/api/track-user/route.ts
@@ -12,15 +12,13 @@ export async function POST(req: Request) {
 
   const rateLimitKey = ip === 'unknown' ? 'unknown-client' : ip;
 
-  if (ip !== '127.0.0.1') {
-    const rateLimitResult = await trackUserRateLimiter.checkWithResult(rateLimitKey);
+  const rateLimitResult = await trackUserRateLimiter.checkWithResult(rateLimitKey);
 
-    if (!rateLimitResult.success) {
-      return NextResponse.json(
-        { success: false, error: 'Too many requests, please try again later.' },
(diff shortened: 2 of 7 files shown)


Aamod007 (approved):
Thanks for the contribution. I went through the changes and the overall approach looks good. This fix addresses the issue effectively and prevents the edge case from occurring in the future. Happy to approve this.

PR #5212, opened 2026-06-11
fix(api): clamp and sanitize grace param instead of rejecting invalid…

### LINKED ISSUE #5202: bug: grace parameter accepts out-of-range values silently — streak count becomes inconsistent without user feedback
### Description

The grace URL parameter (used to set how many days of inactivity are forgiven before a streak resets) is documented to accept values in the range 0–7, with values outside this range being "clamped automatically." However, when a user passes an invalid value such as grace=-5, grace=999, or a non-numeric string like grace=abc, the API silently clamps or ignores the value without returning any warning, error header, or indication in the SVG output. This means users who accidentally pass an invalid grace value receive a badge that reflects different streak logic than they intended, with no way to detect the issue from the response alone.

### Steps to Reproduce

1. Open any browser or use curl to request the following URLs:

   https://commitpulse.vercel.app/api/streak?user=aisshwaryaa8-collab&grace=-1
   https://commitpulse.vercel.app/api/streak?user=aisshwaryaa8-collab&grace=100
   https://commitpulse.vercel.app/api/streak?user=aisshwaryaa8-collab&grace=abc

2. Observe that all three requests return a valid SVG badge with no indication of an invalid parameter.

### Expected Behavior

When a grace value outside the valid range (0–7) is passed, the API should still render the badge (graceful degradation with clamping is acceptable), but it should also communicate the applied value back to the caller — for example, via a response header such as X-CommitPulse-Grace-Applied: 7 or a small visible indicator in the SVG.
When a non-numeric value like grace=abc is passed, the API should either return an appropriate error response (e.g. 400 Bad Request with a JSON body) or fall back to the default value and surface that fallback clearly.
At minimum, there should be consistency: users should not have to guess whether their parameter was accepted, clamped, or silently ignored.

### Screenshots / Logs

_No response_

aisshwaryaa8-collab

### Environment

Chrome

aisshwaryaa8-collab:
… values

## Description

The `grace` parameter was returning a 400 error for out-of-range or 
non-numeric values. This PR makes it silently clamp instead:
- `grace=abc` → falls back to default (1)
- `grace=999` → clamped to 7
- `grace=-5` → clamped to 0

Fixes #5202 

## Pillar

- [x ] 🛠️ Other (Bug fix, refactoring, docs)

## Visual Preview

image

- [ x] I have read the `CONTRIBUTING.md` file.
- [ x] I have tested these changes locally (`localhost:3000/api/streak?user=YOUR_USERNAME`).
- [ x] I have run `npm run format` and `npm run lint` locally and resolved all errors (CI will fail otherwise).
- [ x] My commits follow the Conventional Commits format (e.g., `feat(themes): ...`, `fix(calculate): ...`).
- [x ] I have updated `README.md` if I added a new theme or URL parameter.
- [ x] I have started the repo.
- [x ] I have made sure that i have only one commit to merge in this PR.
- [x ] The SVG output matches the CommitPulse "premium quality" aesthetic standard (no raw elements, smooth animations, correct fonts).
- [ x] (Recommended) I joined the CommitPulse Discord community for contributor discussions, mentorship, and faster PR support.


Aamod007 (approved):
Thanks for the contribution. I went through the changes and the overall approach looks good. This fix addresses the issue effectively and prevents the edge case from occurring in the future. Looks good to merge.

PR #5287, opened 2026-06-11
perf: Compress in-memory SVG cache buffers using Brotli

### LINKED ISSUE #5238: perf: Compress in-memory SVG cache buffers using Brotli
- **Component**: Caching Layer
- **Target File**: [lib/cache.ts](file:///e:/Projects/commitpulse/lib/cache.ts)
- **Problem**: 
  Generated SVGs can be up to 100KB each. Storing thousands of them in-memory as raw strings inside `DistributedCache` creates significant heap memory pressure, leading to garbage collection pauses or serverless function OOMs.
- **Solution**:
  Use Node's built-in `zlib` to compress cached SVGs:
  1. In `DistributedCache.set`, compress string values using `brotliCompressSync`.
  2. In `DistributedCache.get`, check if the stored value is a Buffer, and decompress it using `brotliDecompressSync`.
  3. Ensure a silent fallback to raw string storage if compression fails.
- **Verification**:
  Verify cache integrity by storing a large SVG, asserting that the retrieved value is identical, and checking that the stored memory size is reduced by >70%.

---

MasterJi27:
Resolves #5238. Transparently compresses and decompresses large string/object values stored in TTLCache using Node's zlib Brotli methods.


Aamod007 (approved):
Appreciate you taking the time to submit this PR. I went through the implementation and the approach looks solid.

I'm happy to approve this. Great job!

PR #5288, opened 2026-06-11
perf: Implement Next.js Edge Timeout Racer with Stale Cache fallback

### LINKED ISSUE #5240: perf: Implement Next.js Edge Timeout Racer with Stale Cache fallback
- **Component**: Next.js API Routes
- **Problem**: 
  The GitHub GraphQL request timeout is set to 8s. Vercel edge functions timeout at 10s. If the API is slow, the edge route will time out, returning a raw server error page instead of a friendly fallback.
- **Solution**:
  Implement a timeout race logic:
  1. Introduce a `Promise.race` inside `fetchGitHubContributions` comparing the active fetch against a 4s timeout delay.
  2. If the 4s delay resolves first, check if there is a stale cache entry available.
  3. If available, return the stale cache with `isOfflineFallback: true` rather than waiting for the timeout.
- **Verification**:
  Mock a slow GitHub API (delaying 6s) and assert that the endpoint responds in under 4.5s with cached stale data.

---

MasterJi27:
Resolves #5240. Implements a 4-second timeout race using Promise.race, returning stale cached data with isOfflineFallback: true on timeout or fetch failures.


Aamod007 (approved):
Thanks for putting this together. The code looks clean and solves the issue effectively.

I'm happy to approve this. Great job!

PR #5289, opened 2026-06-11
perf: Implement reset-aware, lock-free token scheduling algorithm

### LINKED ISSUE #5241: perf: Implement reset-aware, lock-free token scheduling algorithm
- **Component**: GitHub Service
- **Problem**: 
  The current token rotation uses a round-robin that selects tokens sequentially even if they are currently rate-limited, causing unnecessary request errors and retries.
- **Solution**:
  Optimize `getGitHubToken()` to use rate-limit telemetry:
  1. Maintain a map of token indexes to their known rate-limit reset epochs (`rateLimitedTokens`).
  2. When requesting a token, filter the list of tokens to only include those whose reset times are in the past.
  3. Select the token with the highest remaining quota or the earliest reset epoch.
- **Verification**:
  Inject a mock token that is rate-limited and verify that rotation immediately bypasses it and proceeds to an active token.

---

MasterJi27:
Resolves #5241. Tracks remaining quota and reset epochs per-token in tokenStats. Schedules the token with the highest remaining quota dynamically, and trips the global circuit breaker to the earliest reset time if all tokens are exhausted.

### DIFF
--- a/lib/github.rotation.test.ts
+++ b/lib/github.rotation.test.ts
@@ -1,5 +1,11 @@
 import { describe, it, expect, beforeEach, afterEach, vi, type Mock } from 'vitest';
-import { fetchWithRetry, getGitHubTokens, clearGitHubApiCacheForTests } from './github';
+import {
+  fetchWithRetry,
+  getGitHubTokens,
+  clearGitHubApiCacheForTests,
+  getTokenStatsForTests,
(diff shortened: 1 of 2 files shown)


Aamod007 (approved):
Thanks for the contribution. I've taken a look at the changes and everything seems to align with what we need here.

Everything looks good. Approving this now.

PR #5338, opened 2026-06-12
bug: Align historical year ranges to user timezone offsets

### LINKED ISSUE #5259: bug: Align historical year ranges to user timezone offsets
- **Component**: GitHub API Query
- **Label**: `bug`, `timezone`
- **Problem**: 
  When requesting historical years (e.g., `year=2023`), the GraphQL API query range is set to UTC, cutting off early morning or late night contributions for users in offset timezones.
- **Solution**:
  1. Construct the from/to ISO dates based on the user's timezone offset:
     `from = YYYY-01-01T00:00:00` in local time.
  2. Convert this local timestamp to UTC before sending the request.
- **Verification**:
  For timezone `UTC-10`, check that the query's start time starts at `2023-01-01T10:00:00Z`.

---

MasterJi27:
Resolves #5259. Aligns the query bounds from and to parameter values with timezone offset boundaries by converting local midnight and local end-of-day timestamps to UTC when fetching historical year contribution calendars.

### DIFF
--- a/app/api/wrapped/route.ts
+++ b/app/api/wrapped/route.ts
@@ -53,6 +53,7 @@ export async function GET(request: Request) {
       hide_background,
       width,
       height,
+      tz,
     } = parseResult.data;
 
     const year = customYear || new Date().getFullYear().toString();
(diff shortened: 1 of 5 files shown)


Aamod007 (approved):
Great work! Timezone offsets are notoriously tricky to handle without heavy external libraries, but your use of Intl.DateTimeFormat to parse parts and calculate the offset milliseconds is very clever and lightweight. 

The tests verifying the offset against Honolulu timezone perfectly validate the boundaries.

**Labels Applied:**
- level:intermediate: Timezone offset calculation using Intl parsing is moderately complex.
- type:bug: Fixes issue #5259 with year bounds.
- quality:clean: Clean, native implementation without adding heavy dependencies, and includes tests.
- gssoc:approved
- mentor:Aamod007

Issue #5264, opened 2026-06-11
bug: Escape all special XML characters to prevent SVG attribute injection


MasterJi27:
- **Component**: Security
- **Target File**: [lib/svg/sanitizer.ts](file:///e:/Projects/commitpulse/lib/svg/sanitizer.ts)
- **Label**: `security`, `bug`
- **Problem**: 
  The custom `label` parameter allows users to inject arbitrary text. If they pass single quotes, double quotes, or backticks, they can break out of SVG attributes and trigger XML layout injection.
- **Solution**:
  1. Refactor `escapeSVGText` to fully escape all unsafe XML characters:
     - `"` -> `&quot;`
     - `'` -> `&apos;`
     - `` ` `` -> `&#96;`
- **Verification**:
  Pass `label="onload='alert(1)'"` and verify that the output SVG renders the string literally and does not contain raw quotes.

---

PR #5395, opened 2026-06-12
bug: escape backticks in svg text

saurabhhhcodes:
Escapes backticks in SVG/XML text output so user-controlled strings cannot break out of attributes.\n\nValidation:\n- npm test -- --run lib/svg/generator.escapeXML.test.ts lib/svg/generator.test.ts\n\nCloses #5264

### DIFF
--- a/lib/svg/generator.escapeXML.test.ts
+++ b/lib/svg/generator.escapeXML.test.ts
@@ -16,6 +16,10 @@ describe('escapeXML', () => {
     expect(escapeXML('hello world')).toBe('hello world');
   });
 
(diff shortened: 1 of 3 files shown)


Aamod007 (approved):
Solid escaping! I went through the changes and ensuring backticks are safely escaped inside SVG text nodes prevents malformed XML strings from breaking the image parser.
**Labels applied:**
- level:intermediate: String escaping logic.
- quality:clean: Protects the XML generator.
- 	ype:bug: Fixes rendering crashes.

PR #5453, opened 2026-06-12
feat(api): implement circuit breaker pattern for GitHub API requests

### LINKED ISSUE #5169: feat(api): implement circuit breaker pattern for GitHub API requests

Currently, the application makes direct calls to the GitHub API without any resilience patterns. If the GitHub API returns a 403 (Rate Limited) or 503 (Service Unavailable), the application continues to attempt requests, which exacerbates the issue, wastes system resources, and can lead to cascading failures across the backend.

I propose implementing a Circuit Breaker pattern. If the GitHub API fails consecutively for a defined threshold (e.g., 5 times), the circuit "opens." While the circuit is open, subsequent requests are immediately rejected or served from the cache without hitting the network, preventing unnecessary API load. After a "cool-down" period, the circuit enters a "half-open" state to test if the service is healthy again. This can be implemented using a library like opossum or via custom middleware.

I considered simple exponential backoff, but that only helps with transient spikes and doesn't protect the system during prolonged service outages. The Circuit Breaker pattern is superior because it provides a "fail-fast" mechanism that protects both our backend and the external GitHub service.

ShreyasPatil3105:
## Summary
Implements a Circuit Breaker pattern using `opossum` in `lib/github.ts` to wrap external GitHub API calls.

## Changes
- Added CircuitBreaker with timeout, error threshold, and reset timeout
- Wrapped all fetchWithRetry calls in apiBreaker.fire()
- Added event logging for circuit breaker state changes
- Added test helper functions for token rotation stats

## Benefits
- Prevents cascading failures during GitHub API outages
- Fails fast when circuit is open
- Automatically recovers after timeout period

Fixes #5169


Aamod007 (changes requested):
This PR is currently marked with the \status:blocked\ label. Please resolve the blockers so we can proceed with a full review and approval.

Aamod007 (changes requested):
Thanks for the contribution! I went through the changes and have evaluated them according to the rubric.

Aamod007 (changes requested):
Please fix the issues that caused the blocked label before this can be approved.

Aamod007 (changes requested):
This PR is currently blocked due to a failing CI check or other blocking issues. Please fix the blocking issues so we can proceed with the review and approval process.

Aamod007 (changes requested):
The circuit-breaker refactor is a big step, but I can’t approve it in this state. lib/github.ts now replaces a lot of the existing retry and GraphQL flow, and the test file changes show several skips in lib/github.test.ts. Please restore or replace the skipped coverage, then verify the new breaker behavior on the same paths that previously handled RATE_LIMITED and token exhaustion.

Aamod007 (changes requested):
The circuit-breaker refactor is still too broad to approve in this state. Please clear the blocker and restore or replace the skipped coverage in lib/github.test.ts and lib/github.rotation.test.ts so the new breaker behavior and the old rate-limit paths are both exercised cleanly.

Aamod007 (changes requested):
This refactor is too broad to approve as-is. lib/github.ts now swaps in opossum, removes etchStarredRepos, and changes the getFullDashboardData payload shape, while lib/github.test.ts has several it.skip(...) cases that cover body-level rate limiting, org aggregation, and timezone handling. Please restore the skipped regression coverage and keep the dashboard contract intact before asking for approval.

Aamod007 (approved):
**Difficulty: advanced** – Implements circuit breaker pattern for GitHub API requests.

**Quality: clean** – Resilience pattern.

**Type: feature** – Error handling.

Excellent pattern!

PR #5514, opened 2026-06-13
fix(stats): report accurate X-Cache-Status from data fetch layer

### LINKED ISSUE #5503: bug: Stats endpoint labels cold non-refresh requests as cache hits
### Bug Description

The JSON stats endpoint reports `X-Cache-Status: HIT` for every non-refresh request, even when the route just fetched data from GitHub/cache service because no cached response was available.

### Evidence

- `app/api/stats/route.ts:125-128` calls `fetchGitHubContributions(user, { bypassCache: shouldBypassCache })`.
- `app/api/stats/route.ts:139` sets `X-Cache-Status` only from `shouldBypassCache`: `MISS` for refresh, `HIT` for everything else.
- There is no signal from `fetchGitHubContributions` used to distinguish a real cache hit from a normal cache miss.

### Impact

Clients, tests, and operators get misleading cache telemetry. A cold request that hits GitHub upstream is labeled as a cache hit, hiding cache misses and making refresh/quota debugging inaccurate.

### Expected Behavior

Return cache metadata from the GitHub data fetch layer, or remove the `X-Cache-Status` header unless the route can accurately report HIT/MISS.

vipul674:
Fixes #5503

## Changes

1. Added `cacheStatus` field to `ExtendedContributionData` interface (`types/index.ts`)
2. Set `cacheStatus` in all return paths of `fetchGitHubContributions` (`lib/github.ts`):
   - HIT for served-from-cache paths
   - MISS for fresh-fetch paths
   - STALE for stale/offline fallback paths
3. Used `userData.cacheStatus` in stats route (`app/api/stats/route.ts`) instead of deriving from `shouldBypassCache` only


Aamod007 (changes requested):
Please fix the issues that caused the blocked label before this can be approved.

Aamod007 (changes requested):
This PR is currently blocked due to a failing CI check or other blocking issues. Please fix the blocking issues so we can proceed with the review and approval process.

Aamod007 (changes requested):
Thanks for wiring cacheStatus through the stats path. This still needs a little polish before merge: etchGitHubContributions now mixes HIT, MISS, and STALE correctly, but the fallback path still depends on the surrounding cache behavior staying perfectly aligned. Please add a small test that covers each status branch in lib/github.ts and the X-Cache-Status header in pp/api/stats/route.ts, then I’m happy to take another look.

Aamod007 (changes requested):
The cache-status wiring is sensible, but I can’t approve it yet because it still needs a focused test pass against the current stats flow. Please re-run the lib/github.ts and pp/api/stats/route.ts branches together so the HIT/MISS/STALE behavior is verified on the live implementation.

Aamod007 (changes requested):
The cache-status wiring in pp/api/stats/route.ts and lib/github.ts is close, but the current 	rackingLoad flag is mutable shared state around getOrSet, so the returned status can become misleading once concurrent requests coalesce. Please move the freshness signal into the value returned by the loader itself or another path that cannot be flipped by overlapping calls.

Aamod007 (approved):
**Difficulty: beginner** – Reports accurate X-Cache-Status from data fetch layer.

**Quality: clean** – Cache reporting fix.

**Type: bug** – Status accuracy.

Good fix!

PR #5515, opened 2026-06-13
fix(stats): record refresh cooldown only after successful fetch

### LINKED ISSUE #5504: bug: Failed stats refresh attempts still consume refresh cooldown
### Bug Description

Manual refresh cooldown is recorded before the GitHub fetch succeeds. If the upstream fetch fails, the user is still placed on cooldown and the next refresh request is served from cache instead of allowing a retry.

### Evidence

- `app/api/stats/route.ts:111-122` checks refresh cooldown and calls `refreshPolicy.recordRefresh(user)` before fetching data.
- `app/api/stats/route.ts:125-128` performs the GitHub fetch afterwards.
- If the fetch throws and the route returns an error, the refresh attempt has still consumed the per-user cooldown window.

### Impact

A transient GitHub/network failure can block a user from retrying a manual refresh for the cooldown period. The UI can then keep serving stale cached data even though the previous refresh did not actually update anything.

### Expected Behavior

Record the refresh cooldown only after a successful fresh fetch, or roll back the cooldown when the upstream refresh fails.

vipul674:
Fixes #5504

### DIFF
--- a/app/api/stats/route.ts
+++ b/app/api/stats/route.ts
@@ -117,13 +117,16 @@ export async function GET(request: Request) {
         remainingMs: refreshPolicy.getRemainingCooldown(user),
       });
       shouldBypassCache = false;
-    } else {
-      refreshPolicy.recordRefresh(user);
     }
   }
 
   try {
     const userData = await fetchGitHubContributions(user, { bypassCache: shouldBypassCache });
+
+    if (shouldBypassCache) {
+      refreshPolicy.recordRefresh(user);
+    }
+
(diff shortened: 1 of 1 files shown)


Aamod007 (approved):
Good logic correction! I went through the changes and recording the refresh cooldown *only* after a successful fetch prevents users from being locked out of retries when an API fails.
**Labels applied:**
- level:intermediate: State management logic.
- quality:clean: Correctly repositions the cache setter.
- 	ype:bug: Fixes UX retry lockouts.

PR #5516, opened 2026-06-13
fix(resume): return 500 when MongoDB missing in production

### LINKED ISSUE #5505: bug: Resume confirmation reports success when MongoDB is not configured
### Bug Description

The resume confirmation endpoint returns success when `MONGODB_URI` is missing, even though the profile is not saved.

### Evidence

- `app/api/student/resume/confirm/route.ts:82-86` checks `!process.env.MONGODB_URI`, logs a warning, and returns `{ success: true, bypassed: true }`.
- This path is not limited to local development or tests.
- The user receives a successful response although no `StudentProfile.findOneAndUpdate` call is made.

### Impact

If the deployment is missing database configuration, resume confirmation appears to work but silently drops the user profile. The next page/load will not have the confirmed resume data, which is a confusing data-loss failure mode.

### Expected Behavior

Only bypass persistence in explicit local/test environments. In production or preview deployments, return a 500 configuration error when MongoDB is unavailable so the client can show a real failure.

vipul674:
## What does this PR do?

Fixes resume confirmation endpoint to return a 500 error in production/preview when MONGODB_URI is not configured, instead of silently returning success and dropping the profile save.

## Related issue

Fixes #5505

## Changes Made

- Only bypass persistence in local development (when NODE_ENV is not production and VERCEL_ENV is not production/preview)
- In production/preview deployments, return 500 with "Server configuration error: Database not configured"

## Checklist

- [x] My PR has a linked issue (Fixes #5505)
- [x] I have pulled the latest `main` and resolved any conflicts

### DIFF
--- a/app/api/student/resume/confirm/route.ts
+++ b/app/api/student/resume/confirm/route.ts
@@ -81,7 +81,17 @@ export async function POST(req: Request) {
 
   try {
     if (!process.env.MONGODB_URI) {
-      console.warn('MONGODB_URI is not set. Bypassing student profile save.');
+      const isProduction =
+        process.env.NODE_ENV === 'production' ||
+        process.env.VERCEL_ENV === 'production' ||
+        process.env.VERCEL_ENV === 'preview';
+      if (isProduction) {
+        return NextResponse.json(
+          { success: false, error: 'Server configuration error: Database not configured' },
+          { status: 500 }
+        );
+      }
+      console.warn('MONGODB_URI is not set. Bypassing student profile save (local dev).');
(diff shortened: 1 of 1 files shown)


Aamod007 (approved):
Solid edge case fix! I went through the changes and explicitly returning a 500 when MongoDB connects fail prevents hanging requests in the resume generation pipeline.
**Labels applied:**
- level:intermediate: Database connection error handling.
- quality:clean: Direct and effective fix.
- 	ype:bug: Protects production runtime.

PR #5695, opened 2026-06-14
fix(useLocalStorage): use lazy useState initializer to eliminate stale initialValue on first render

### LINKED ISSUE #5437: [Bug] useLocalStorage hook reads from window.localStorage in useEffect but setValue writes synchronously — inconsistent read/write timing causes stale state on first render
### Description

`hooks/useLocalStorage.ts` reads from `window.localStorage` inside a `useEffect` (deferred to after mount) but `setValue` writes to `window.localStorage` synchronously:

```ts
// Read: deferred to after mount (SSR-safe)
useEffect(() => {
  const item = window.localStorage.getItem(key);
  if (item !== null) setStoredValue(JSON.parse(item));
}, [key]);

// Write: synchronous, no SSR guard
const setValue = (value: T): void => {
  setStoredValue(value);
  window.localStorage.setItem(key, JSON.stringify(value));
};
```

On the first render, `storedValue` is always `initialValue` even if localStorage has a saved value — the deferred `useEffect` hasn't run yet. If `setValue` is called before the `useEffect` fires (e.g. from another `useEffect` with `[]`), it overwrites the stored value with `initialValue` first render state, losing previously saved data.

### Steps to Reproduce

1. Save a value to localStorage via `setValue`
2. Reload the page
3. On first render, `storedValue` is `initialValue` — the saved value appears only after hydration

### Expected Behavior

The hook should initialize state from localStorage synchronously using a lazy initializer (`useState(() => { ... })`) with an SSR guard, consistent with `useRecentSearches.ts` which handles this correctly.

### Screenshots / Logs

N/A

N/A

### Environment

Chrome

KaparthyReddy:
## Description
Fixes #5437

`hooks/useLocalStorage.ts` read from `window.localStorage` inside a deferred `useEffect` — on the first render `storedValue` was always `initialValue` even when localStorage had a saved value. If `setValue` was called before the deferred effect fired, it overwrote previously saved data:

```ts
// Before — deferred read causes stale first render
const [storedValue, setStoredValue] = useState<T>(initialValue);
useEffect(() => {
  const item = window.localStorage.getItem(key);
  if (item !== null) setStoredValue(JSON.parse(item));
}, [key]);
```

Fixed by using a lazy `useState` initializer that reads from localStorage synchronously with an SSR guard — consistent with the pattern already used in `useRecentSearches.ts`:

```ts
// After — synchronous read on first render, SSR-safe
function readFromStorage<T>(key: string, initialValue: T): T {
  if (typeof window === 'undefined') return initialValue;
  try {
    const item = window.localStorage.getItem(key);
    return item !== null ? JSON.parse(item) : initialValue;
  } catch { return initialValue; }
}

const [storedValue, setStoredValue] = useState<T>(() => readFromStorage(key, initialValue));
```

Changes made:
- Extracted `readFromStorage` helper with SSR guard
- Replaced `useState(initialValue)` + deferred `useEffect` read with `useState(() => readFromStorage(key, initialValue))`
- Kept a `useEffect` to re-sync when the `key` prop changes
- `setValue` is unchanged — writes remain synchronous
- Verified zero new TypeScript errors introduced by this change

## Pillar

## Visual Preview
N/A — hook behaviour fix with no visual output.

- [x] The SVG output matches the CommitPulse "premium quality" aesthetic standard.
- [ ] (Recommended) I joined the CommitPulse Discord community.

### DIFF
--- a/hooks/useLocalStorage.ts
+++ b/hooks/useLocalStorage.ts
@@ -1,21 +1,37 @@
 'use client';
-
 import { useEffect, useState } from 'react';
 
+/**
+ * Reads from localStorage synchronously with an SSR guard.
+ * Used as a lazy useState initializer so the stored value is available
+ * on the first render — consistent with useRecentSearches.ts.
+ * Returns initialValue when running on the server or when the key is absent.
+ */
+function readFromStorage<T>(key: string, initialValue: T): T {
+  if (typeof window === 'undefined') return initialValue;
+  try {
+    const item = window.localStorage.getItem(key);
+    return item !== null ? (JSON.parse(item) as T) : initialValue;
+  } catch {
+    return initialValue;
+  }
+}
+
 export function useLocalStorage<T>(key: string, initialValue: T): readonly [T, (value: T) => void] {
-  const [storedValue, setStoredValue] = useState<T>(initialValue);
+  // Lazy initializer reads from localStorage synchronously on first render
+  // instead of deferring to a useEffect — eliminates the stale initialValue
+  // flash on mount and prevents setValue from overwriting stored data before
+  // the deferred read had a chance to run.
+  const [storedValue, setStoredValue] = useState<T>(() => readFromStorage(key, initialValue));
 
+  // Re-sync when the key changes (e.g. component reused with a different key).
+  // setState inside an effect is intentional here — we are synchronising React
+  // state with an external system (localStorage) when the key prop changes,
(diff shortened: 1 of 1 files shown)


Aamod007 (approved):
Nice bug fix. 
eadFromStorage() in hooks/useLocalStorage.ts moves the storage read into a lazy initializer, which fixes the stale first render and keeps the key-change resync behavior explicit in the effect below it.

PR #5715, opened 2026-06-14
fix(svg): replace FNV-1a with MurmurHash3 for deterministic constellation seeding

### LINKED ISSUE #5252: feat: Seed constellation particle generation using MurmurHash3
- **Component**: Generative Particles
- **Target File**: [lib/svg/constellation.ts](file:///e:/Projects/commitpulse/lib/svg/constellation.ts)
- **Label**: `enhancement`, `geometry`
- **Problem**: 
  The constellation background uses a simple character-adding seed, meaning usernames like "ab" and "ba" generate identical star patterns.
- **Solution**:
  1. Add a MurmurHash3 hashing function in `lib/svg/constellation.ts`.
  2. Generate a 32-bit integer hash from the username string.
  3. Use this hash as the seed for the deterministic particle randomizer.
- **Verification**:
  Assert that users "ab" and "ba" yield different random coordinates for their constellation stars.

---

MasterJi27:
## Summary

The deterministicRandom function in lib/svg/generator.ts used FNV-1a hashing, which has poor collision resistance for short anagrammatic strings like usernames. Two users with anagrammatic usernames (e.g. 'ab' and 'ba') could generate identical constellation star patterns.

Replace FNV-1a with MurmurHash3, a non-cryptographic hash with better avalanche properties, ensuring distinct particle layouts for distinct user/year combinations.

## Changes

- **lib/svg/generator.ts**: Replaced FNV-1a implementation in deterministicRandom with MurmurHash3 32-bit. The function signature and return type are unchanged — this is a drop-in replacement that improves hash distribution.

## Testing

- All 178 generator.test.ts tests pass
- All 9 constellation.test.ts tests pass

## Issue

Fixes #5252

### DIFF
--- a/lib/svg/generator.deterministicRandom.test.ts
+++ b/lib/svg/generator.deterministicRandom.test.ts
@@ -3,7 +3,6 @@ import { deterministicRandom } from './generator';
 
 describe('Helper Function: deterministicRandom', () => {
   it('Test 1: should return deterministic values for standard seeds', () => {
-    // FNV-1a produces completely deterministic hashes based on identical strings.
     const run1 = deterministicRandom('commitpulse-seed');
     const run2 = deterministicRandom('commitpulse-seed');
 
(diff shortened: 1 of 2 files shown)


Aamod007 (approved):
This is a good deterministic randomness upgrade. lib/svg/generator.ts swaps the hash implementation while keeping the signature stable, and the updated generator test pins the new output, so the change is well covered.

PR #5716, opened 2026-06-14
fix(svg): apply Painter's Algorithm sorting for overlapping contribution towers

### LINKED ISSUE #5247: bug: Implement strict Painter's Algorithm sorting for overlapping towers
- **Component**: Geometry Sorting
- **Target File**: [lib/svg/generator.ts](file:///e:/Projects/commitpulse/lib/svg/generator.ts)
- **Label**: `bug`, `geometry`
- **Problem**: 
  In certain views, taller contribution towers render *behind* shorter towers located in front of them due to a simple index sorting bug, breaking 3D perspective.
- **Solution**:
  1. Implement a topological sort of the rendering grid.
  2. Compute rendering order by comparing projection coordinates: towers furthest from the camera must be drawn first.
  3. Sort the array of rendering coordinates `(x, y)` such that `x + y` determines drawing precedence.
- **Verification**:
  Render a grid where a tower at `(2,2)` is height 10 and `(3,2)` is height 1, and ensure the SVG path tags are ordered correctly.

---

MasterJi27:
## Summary

Towers in the isometric contribution graph are rendered in column-major iteration order (week index ascending, then day-of-week ascending). This causes shorter towers to appear in front of taller towers when they overlap, because SVG renders later elements on top of earlier ones.

Sort 	owerData by (row + col) ascending before rendering, so that towers further from the viewer (lower screen-space Y) are drawn first and towers closer to the viewer (higher screen-space Y) are drawn later on top — implementing the Painter's Algorithm.

## Changes

- **lib/svg/generator.ts**: Added const sortedTowers = [...towerData].sort((a, b) => (a.row + a.col) - (b.row + b.col)) at the start of 
enderTowers and changed the for loop to iterate sortedTowers instead of 	owerData. This is a shallow copy sort that doesn't mutate the input array.

## Testing

- All 178 generator.test.ts tests pass

## Issue

Fixes #5247

### DIFF
--- a/lib/svg/generator.ts
+++ b/lib/svg/generator.ts
@@ -445,7 +445,9 @@ function renderTowers(
   const dx_sf = rnd(dx * sf);
   const dy_sf = rnd(dy * sf);
 
-  for (const t of towerData) {
+  const sortedTowers = [...towerData].sort((a, b) => a.row + a.col - (b.row + b.col));
+
+  for (const t of sortedTowers) {
     const isGhost = t.isGhost;
     let strokeColor = '';
     let leftRightFillAttr = '';
--- a/lib/svg/layout.accessibility.test.ts
+++ b/lib/svg/layout.accessibility.test.ts
@@ -97,7 +97,7 @@ describe('Layout Accessibility Standards & Screen Reader Aria Compliance', () =>
   });
 
   // 4. Keyboard Control Path & Ordering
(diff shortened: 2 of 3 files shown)


Aamod007 (approved):
Good SVG rendering fix. lib/svg/generator.ts applies a shallow sorted copy before rendering towers, and the paired accessibility tests update the DOM-order expectation to match the Painter’s Algorithm change, so this is a solid geometry/overlap improvement.

PR #5798, opened 2026-06-15
perf: optimize Zod schema validation using LRU cache

### LINKED ISSUE #5244: perf: Optimize Zod Schema validations using compilation caches
- **Component**: Parameter Validation
- **Problem**: 
  `streakParamsSchema.safeParse` compiles and parses parameters on every API hit. In high-traffic scenarios, this results in noticeable CPU time spent in parsing loops.
- **Solution**:
  Implement validation caching:
  1. Create a lightweight LRU validation cache: `const validationCache = new Map<string, any>();`
  2. Stringify the incoming URL search query and check the cache.
  3. If present, return the parsed object. Otherwise, validate with Zod and cache the result.
- **Verification**:
  Run a benchmark script requesting identical params and assert a CPU time reduction.

---

MasterJi27:
## Summary

\streakParamsSchema.safeParse\ compiles and parses parameters on every API hit. In high-traffic scenarios, this results in noticeable CPU time spent in parsing loops.

## Changes

- **\pp/api/streak/route.ts\**: Added a lightweight Map-based LRU validation cache (max 256 entries) keyed by the serialized URL search params. Repeated identical requests skip the full Zod parse pipeline and return the cached result.

## Testing

- Existing streak route tests pass
- Cache correctly returns cached results for identical params

## Issue

Fixes #5244

### DIFF
--- a/app/api/streak/route.ts
+++ b/app/api/streak/route.ts
@@ -33,6 +33,24 @@ import { quotaMonitor } from '@/services/github/quota-monitor';
 import { refreshPolicy } from '@/services/github/refresh-policy';
 import { refreshRateLimiter } from '@/services/github/refresh-rate-limiter';
 
+const VALIDATION_CACHE_MAX = 256;
+const validationCache = new Map<string, ReturnType<typeof streakParamsSchema.safeParse>>();
+
+function cachedValidation(
+  key: string,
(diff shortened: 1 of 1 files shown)


Aamod007 (approved):
Excellent perf optimization

PR #5929, opened 2026-06-17
fix: increase RateLimiter maxSize to prevent IP eviction (#5893)

### LINKED ISSUE #5893: [Bug]: RateLimiter maxSize=1 disables per-IP rate limiting in memory-fallback mode
## Bug Description

In `app/api/ci-analytics/route.ts:10` and `app/api/pr-insights/route.ts:10`, the RateLimiter is constructed with `maxSize=1`:

```typescript
const ciAnalyticsLimiter = new RateLimiter(10, 60_000, 1);
```

The third argument (`maxSize=1`) is forwarded to `TTLCache`, limiting the cache to **one entry total**. When IP A hits, entry for IP A is stored. When IP B arrives, IP A is evicted, and IP B gets a fresh window. Under concurrent IPs, every request gets a fresh rate-limit window — rate limiting is completely defeated without KV/Redis.

Same bug in `prInsightsLimiter`.

## Expected Behavior
`maxSize` should be set to something reasonable like `1000` to store separate rate-limit entries per distinct IP, or use an unbounded cache.

## Severity
Medium — rate limiting ineffective under multi-IP traffic.

vipul674:
Fixes #5893

## Problem
RateLimiter was instantiated with `maxSize=1`, allowing only one IP to be tracked at a time. When a new IP made a request, the previous IP's entry was evicted, effectively disabling rate limiting for all users except the most recent one.

## Fix
Increased `maxSize` from `1` to `10_000` in both ci-analytics and pr-insights route rate limiters, matching the default used elsewhere in the codebase.


Aamod007 (approved):
**Difficulty: beginner** – Increases RateLimiter maxSize from 1 to 10,000 in ci-analytics and pr-insights routes to prevent IP eviction.

**Quality: clean** – Simple config fix.

**Type: bug** – Fixes issue #5893, prevents premature IP eviction.

Good fix!

PR #5932, opened 2026-06-17
fix: prevent rate limiter quota leak on limit exceeded (#5895)

### LINKED ISSUE #5895: [Bug]: Refresh rate limiter consumes quota on cooldown-blocked refreshes
## Bug Description

In `app/api/github/route.ts:64-80`, the refresh flow has a quota leak:

1. `refreshRateLimiter.checkLimit(ip)` — increments the rate-limit counter
2. `refreshPolicy.isRefreshAllowed(username)` — checks cooldown, may reject

When step 2 rejects (user is in cooldown), the quota slot consumed in step 1 is **never returned**. The user exhausts their 3 refreshes/hour by hitting the cooldown wall without performing a single actual refresh.

## Expected Behavior
Check the refresh policy before consuming the rate-limit quota. Or implement a reservation pattern that refunds unused quota.

## Severity
Medium — users lose refresh quota to cooldown rejections.

vipul674:
Fixes #5895

## Problem
When using Upstash Redis/KV, the rate limiter used `INCR` to increment the counter before checking if the limit was exceeded. This meant rejected requests still consumed quota, effectively lowering the rate limit below the configured value.

For example, with `limit=10`, if a user makes 10 requests (counter=10), the 11th request would:
1. INCR → counter becomes 11
2. Check: 11 > 10 → reject

But now the counter is 11, meaning the user is permanently one request over the limit until the window resets.

## Fix
Check the current count first using `GET` + `TTL`, and only `INCR` if the count is below the limit. Rejected requests no longer consume quota.


Aamod007 (approved):
**Difficulty: intermediate** – Changes rate-limit.ts to GET+check before INCR, preventing quota leak when limit exceeded (39 additions, 12 deletions).

**Quality: clean** – Fixes INCR-on-rejection bug.

**Type: bug + security** – Prevents rate limiter bypass issue #5895.

Important security fix!

PR #5933, opened 2026-06-17
fix: remove double cache write in contributions fetch (#5896)

### LINKED ISSUE #5896: [Bug]: fetchContributionsUncached double-writes cache with unprocessed data — race window returns stale data
## Bug Description

In `lib/github.ts:450-460` and `lib/github.ts:535-538`, there is a double-write to the same cache key:

```typescript
// Inside fetchContributionsUncached (line 450):
if (!options.bypassCache) {
    await contributionsCache.set(key, { calendar, ... }, LONG_CACHE_TTL);  // unprocessed weeks
}
const processedWeeks = mapWeeks(calendar.weeks);  // adds locAdditions/locDeletions
return { calendar: { ...calendar, weeks: processedWeeks }, ... };
```

`fetchGitHubContributions` wraps this via `contributionsCache.getOrSet(key, coalescedLoad, ...)`. `getOrSet` ALSO caches the return value via `this.set(key, freshData, ttlMs)`.

Both writes target the same `DistributedCache` instance. Between the inner unprocessed write and the outer processed write, a concurrent request reading the cache receives unprocessed data where `locAdditions` and `locDeletions` are silently missing.

## Expected Behavior
Only write processed data to the cache. Remove the inner `set` call in `fetchContributionsUncached`, or have it write the fully processed object.

## Severity
Low — intermittent missing computed fields in cached data.

vipul674:
Fixes #5896

## Problem
`fetchContributionsUncached` was calling `contributionsCache.set()` to cache the fetched data. Then, `fetchGitHubContributions` called `contributionsCache.getOrSet()` which internally calls `set()` again. This caused redundant writes to both local cache and Redis for every fetch.

## Fix
Removed the explicit `set()` call in `fetchContributionsUncached` since `getOrSet` handles caching. This eliminates the duplicate write and reduces Redis traffic.


Aamod007 (approved):
**Difficulty: intermediate** – Removes duplicate cache write in github.ts fetchContributionsUncached (15 deletions), moves single write to coalescedLoad forceRefresh path.

**Quality: clean** – Eliminates redundant I/O.

**Type: bug** – Fixes double cache write issue #5896.

Clean refactor!

PR #6009, opened 2026-06-18
test: add DST transition boundary tests for calculateStreak

### LINKED ISSUE #5255: bug: Map DST transition boundaries using absolute date strings
- **Component**: Contributions Calculations
- **Target File**: [lib/calculate.ts](file:///e:/Projects/commitpulse/lib/calculate.ts)
- **Label**: `bug`, `timezone`
- **Problem**: 
  During DST spring-forward transitions, a day has only 23 hours. Simple date math (dividing epochs by 86400) shifts subsequent days by an hour, causing contributions to shift into the wrong calendar date cell.
- **Solution**:
  1. Refactor calendar day indexing in `calculate.ts` to map to absolute dates (`YYYY-MM-DD`).
  2. Compare day strings directly instead of dividing milliseconds by 24 hours.
- **Verification**:
  Inject a contribution calendar containing a DST transition date (e.g. March 10) and verify that all day indexes map to correct dates.

---

MasterJi27:
Fixes #5255

### DIFF
--- a/lib/calculate.test.ts
+++ b/lib/calculate.test.ts
@@ -2811,3 +2811,80 @@ describe('normalizeCalendarToTimezone', () => {
     expect(allDays[0].contributionCount).toBe(12);
   });
 });
+
+describe('DST transition boundary handling', () => {
+  it('maps contribution days to absolute date strings during spring-forward DST', () => {
+    const calendar: ContributionCalendar = {
+      totalContributions: 3,
+      weeks: [
+        {
+          contributionDays: [
+            { contributionCount: 1, date: '2024-03-09' },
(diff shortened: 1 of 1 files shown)


Aamod007 (approved):
Great edge case coverage @MasterJi27! The DST transition tests in lib/calculate.test.ts are exactly the kind of timezone gotchas that'd slip through without dedicated tests — spring-forward (America/New_York, March 2024), fall-back (November 2024), and the 'no shift across DST boundary' case at line 2860 all look correct. The assertion on longestStreak: 1 for the non-consecutive March 8-11 gap proves the logic handles boundary-adjacent gaps too. Nice work!

PR #6011, opened 2026-06-18
test: add comprehensive XML validation for SVG generator outputs

### LINKED ISSUE #5280: test: Add XML validation checks for SVG generator outputs
- **Component**: SVG Test Suite
- **Target File**: [lib/svg/generator.test.ts](file:///e:/Projects/commitpulse/lib/svg/generator.test.ts)
- **Label**: `ci`, `test`
- **Problem**: 
  Changes to the SVG templates can produce malformed XML that fails to render.
- **Solution**:
  1. Create a test that parses all generator outputs through an XML parser.
  2. Fail the test if any validation errors are detected.
- **Verification**:
  Inject a missing tag in a template and confirm that the test suite fails.

---

MasterJi27:
Fixes #5280

### DIFF
--- a/lib/svg/generator.test.ts
+++ b/lib/svg/generator.test.ts
@@ -2374,13 +2374,14 @@ describe('XML Validation - All Generator Outputs', () => {
 
   const mockMonthlyStats: MonthlyStats = {
     currentMonthTotal: 42,
+    previousMonthTotal: 30,
+    deltaPercentage: 40,
+    deltaAbsolute: 12,
     currentMonthName: 'June',
-    monthlyStats: [],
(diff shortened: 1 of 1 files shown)


Aamod007 (approved):
Solid work @MasterJi27! The ssertValidSVG helper using DOMParser + parsererror check is exactly what we needed to catch malformed XML at the test level. I especially appreciate that you hit every generator function — generateMonthlySVG, generateHeatmapSVG, generateRateLimitSVG with circuit breaker, even generateVersusSVG — and tossed in unicode/XSS username tests for good measure in lib/svg/generator.test.ts. Merging!

PR #6020, opened 2026-06-19
fix: Badge result can become inconsistent with the latest input

### LINKED ISSUE #3606: bug: Badge result can become inconsistent with the latest input (race condition between image load/error)
### Description

badgeUrl changes with previewUsername. onLoad / onError sets:

setBadgeResult({ username: previewUsername, status: 'loaded' | 'error' })

But image load/error callbacks can arrive out of order when the user changes the username quickly (debounce + instant input). This can set badgeResult for an older URL after the user has already typed a new one.

### Steps to Reproduce

Type a valid username (e.g. torvalds) to trigger badge load.
Quickly replace with another username (e.g. gaearon) before the first badge finishes loading.
Watch badge preview state flicker between loaded/error.

### Expected Behavior

The badge state should correspond strictly to the most recent previewUsername / badgeUrl.
Out-of-order image events should be ignored.

### Screenshots / Logs

_No response_

_No response_

### Environment

Chrome

Aditya8369:
Summary of Work
Identified Race Condition: The onLoad and onError image callbacks inside the 

LandingPageClient.tsx
 badge element would run even if the user had already changed the input. When an older badge image loaded slower than a newer one, the callback would overwrite the correct, newer badgeResult state.
Added Ref Tracking: Created a latestPreviewUsernameRef React Ref inside 

LandingPageClient.tsx
 to track the most recent previewUsername.
Guarded Callbacks: Updated onLoad and onError image event handlers to only trigger state changes if the username at the time of rendering matches the latest value in the ref.
Verified Stability: Ran TypeScript compiler and lint checks which passed successfully with no errors.

closes #3606


Aamod007 (approved):
Solid fix @Aditya8369! The race condition in LandingPageClient.tsx where out-of-order image callbacks could overwrite adgeResult with stale data was a real issue. The latestPreviewUsernameRef at line 370 and the guards on onLoad (line 903) and onError (line 907) are exactly the right pattern. The useEffect hook at line 372 keeps the ref in sync. Clean and focused!

PR #6031, opened 2026-06-19
fix(rate-limit): replace non-atomic get-then-set with atomic incr to prevent TOCTOU race

### LINKED ISSUE #5894: [Bug]: TOCTOU race condition in memory-fallback rate limiter allows concurrent request bypass
## Bug Description

In `lib/rate-limit.ts:97-139` (`checkWithResult`) and `lib/rate-limit.ts:185-215` (`rateLimit`), the rate-limit check uses a non-atomic get-then-set pattern:

```typescript
const record = await this.cache.get(ip);       // (A)
const count = record?.count ?? 0;
if (count >= this.limit) return false;
// ...
await this.cache.set(ip, { count: 1, ... });  // (B)
```

Between (A) and (B), concurrent requests all read `count=0` and all proceed. This allows N concurrent requests to bypass the rate limit simultaneously.

Same pattern exists in the standalone `rateLimit()` function with `trackers` cache. This affects every `RateLimiter` instance when KV (Upstash Redis) is not configured.

## Expected Behavior
Use an atomic increment-and-check operation, or implement a lock/per-mutex for the cache entry.

## Severity
Medium — rate limiting bypass under concurrent requests.

vipul674:
Fixes #5894

## Problem

In `lib/rate-limit.ts`, the rate-limit check uses a non-atomic get-then-set pattern:

```typescript
const record = await this.cache.get(ip);       // (A)
const count = record?.count ?? 0;
if (count >= this.limit) return false;
// ...
await this.cache.set(ip, { count: 1, ... });  // (B)
```

Between (A) and (B), concurrent requests all read `count=0` and all proceed. This allows N concurrent requests to bypass the rate limit simultaneously.

Same pattern exists in the standalone `rateLimit()` function with `trackers` cache.

## Fix

Replaced the non-atomic get-then-set pattern with atomic `incr` calls in both:

1. `RateLimiter.checkWithResult()` — uses `this.cache.incr()`
2. `rateLimit()` function — uses `trackers.incr()`

The `DistributedCache.incr()` method already provides atomic increment behavior:
- **Redis/KV**: Uses Lua script for true atomicity (`INCR` + `EXPIRE`)
- **Memory fallback**: Single-operation increment without read-then-write gap

Also fixed:
- `RateLimiter.reset()` now deletes the correct key (`ratelimit:${ip}`)
- `RateLimiter.remaining()` reads from the correct key format

## Type of Change

- Bug fix

## Checklist

- [x] I have run `npm run format` and `npm run lint` locally.

Closes #5894


Aamod007 (approved):
Solid fix @vipul674! The TOCTOU race in lib/rate-limit.ts was a real issue — the old get-then-set pattern at lines 126-128 could let concurrent requests slip through. Switching to 	his.cache.incr() and 	rackers.incr() is the right approach. Also appreciate the bonus fixes: correcting the key prefix in 
eset() and 
emaining() was overdue. Tests updated cleanly too. Looks good!

PR #6044, opened 2026-06-19
fix: integrate border parameter into Zod validation schema

### LINKED ISSUE #6017: Streak route border parameter is read and sanitized but never used — dead code that bypasses Zod validation
The `border` parameter is read and sanitized in the streak route but never actually passed to the SVG generator, meaning user-provided border values are silently ignored.

**Where:**
`app/api/streak/route.ts:309-310`
```typescript
const borderParam = searchParams.get('border');
const sanitizedBorder = borderParam ? borderParam.replace(/[^a-fA-F0-9]/g, '') : undefined;
```

Then at line 328:
```typescript
border: sanitizedBorder,
```

**The issue:**
1. The `border` parameter is NOT part of the Zod validation schema (`streakParamsSchema` in `lib/validations.ts`). Every other parameter goes through schema validation, but this one is read directly from `searchParams`.
2. The `border` value gets put into the `params` object, but looking at the SVG generator code, this field is either unused or not properly wired up.
3. The sanitization regex `[^a-fA-F0-9]` only allows hex characters, which is overly restrictive for what should be a border specification.

**Why this matters:**
- It bypasses the validation pattern used everywhere else
- Users who set `?border=...` think they're customizing their badge, but it does nothing
- It's dead code that creates confusion for anyone maintaining this

This should either be properly integrated into the Zod schema and SVG generator, or removed entirely to avoid confusion.

atul-upadhyay-7:
## Summary

This PR fixes the issue where the  parameter was read and sanitized in the streak route but never actually passed to the SVG generator, meaning user-provided border values were silently ignored.

## Changes

1. **Added  parameter to Zod validation schema** ():
   - Added proper validation for hex color format (with or without )
   - Used the existing  function for consistent sanitization
   - Applied the same pattern as other color parameters (, , )

2. **Removed manual border reading/sanitization** ():
   - Removed the manual  call
   - Removed the manual regex sanitization that was overly restrictive
   - Added  to the destructured parameters from the validated schema

3. **Used validated border value** ():
   - Updated the  object to use the validated  value from the schema
   - The value is now properly sanitized and validated before being passed to the SVG generator

## Issue

Fixes #6017

## Testing

- All validation tests pass (297 tests)
- All SVG generator tests pass (61 tests)
- Linting passes with no new errors

## Benefits

1. **Consistency**: The  parameter now follows the same validation pattern as other parameters
2. **Proper validation**: Invalid hex colors are now rejected with a clear error message
3. **No dead code**: The parameter is now properly integrated and functional
4. **Better UX**: Users who set  will now actually see their border applied to the badge


Aamod007 (approved):
Glad to see the \order\ parameter finally wired up properly through Zod validation in \lib/validations.ts:504\ and actually passed through to the SVG generator in \pp/api/streak/route.ts:327\. The \sanitizeHexColor\ transform ensures consistent sanitization. Clean fix for a dead-param issue!

Aamod007 (approved):
## Review

Already labeled. Integrates the order parameter into Zod validation schema (pp/api/streak/route.ts:163) and removes the unsafe manual sanitization (:312-313). Clean security improvement — leverages Zod's built-in parsing instead of manual regex sanitization.

Approved! 🚀

PR #6045, opened 2026-06-19
fix: sanitize error messages in architecture endpoint (#6016)

### LINKED ISSUE #6016: Architecture endpoint returns raw error messages that can leak server paths and access tokens
The architecture endpoint returns raw error messages to the client, which can leak internal system details.

**Where:**
`app/api/architecture/route.ts:643`
```typescript
return NextResponse.json({ error: errorMessage }, { status: 500 });
```

**What gets leaked:**
The `catch` block at line 636 catches any exception and returns `error.message` directly. Depending on what fails, this could include:
- Full filesystem paths (e.g., `/tmp/commitpulse-arch-owner-repo-xxxxx`)
- Git clone failure details including the URL with the embedded access token
- Node.js internal error messages
- TypeScript compiler error details with file paths

**Example scenario:**
If the git clone fails for a reason that includes the URL in the error message, the response would contain:
```
Failed to clone https://x-access-token:ghp_xxxx@github.com/owner/repo.git
```

This directly exposes the server's GitHub PAT to any caller.

**Even the "safe" error at line 298 leaks info:**
```typescript
error: 'Failed to clone repository. Make sure the repository exists and is public (or access token is set).'
```
The phrase "or access token is set" confirms to an attacker that the server uses tokens, which is useful info for further attacks.

**Fix:**
Return a generic error message for all 500 responses and only log the details server-side. Something like:
```typescript
return NextResponse.json({ error: 'Failed to analyze repository' }, { status: 500 });
```

atul-upadhyay-7:
## Summary

This PR fixes a security vulnerability where the architecture endpoint returns raw error messages to the client, which can leak internal system details including server paths and access tokens.

## Changes

1. **Removed token info leak from clone error message** ():
   - Changed from: 'Failed to clone repository. Make sure the repository exists and is public (or access token is set).'
   - Changed to: 'Failed to clone repository. Make sure the repository exists and is accessible.'
   - The phrase 'or access token is set' was leaking information that the server uses tokens

2. **Sanitized catch block error response** ():
   - Changed from: Returning raw error.message which could contain server paths, git URLs with embedded tokens, or internal error details
   - Changed to: Returning a generic error message 'Failed to analyze repository. Please try again later.'
   - The detailed error is still logged server-side for debugging

## Security Benefits

- **No more path leaks**: Server filesystem paths are no longer exposed to clients
- **No more token leaks**: Git clone URLs with embedded access tokens are never returned
- **No more internal error details**: Node.js/TypeScript error messages with file paths are hidden
- **Reduced attack surface**: Attackers can no longer confirm token usage via error messages

## Issue

Fixes #6016

## Testing

- Linting passes with no new errors
- Error details are still logged server-side for debugging purposes

### DIFF
--- a/app/api/architecture/route.ts
+++ b/app/api/architecture/route.ts
@@ -294,8 +294,7 @@ export async function POST(req: NextRequest) {
       }
       return NextResponse.json(
         {
-          error:
-            'Failed to clone repository. Make sure the repository exists and is public (or access token is set).',
+          error: 'Failed to clone repository. Make sure the repository exists and is accessible.',
         },
         { status: 404 }
       );
@@ -637,11 +636,10 @@ export async function POST(req: NextRequest) {
     });
   } catch (error) {
     console.error('Architecture visualizer route crashed:', error);
-    const errorMessage =
-      error instanceof Error
-        ? error.message
-        : 'An unexpected error occurred while analyzing the repository.';
-    return NextResponse.json({ error: errorMessage }, { status: 500 });
+    return NextResponse.json(
+      { error: 'Failed to analyze repository. Please try again later.' },
+      { status: 500 }
+    );
   } finally {
     // Clean up temporary directory
     if (tempDir && fs.existsSync(tempDir)) {


Aamod007 (approved):
Good security hardening in \pp/api/architecture/route.ts\ — sanitizing the clone error at line 297 to avoid leaking token info, and replacing the raw \error.message\ return in the catch block at line 639 with a generic message. The detailed errors still get logged server-side, which is the right trade-off. Nice work!

PR #6120, opened 2026-06-20
fix(validations): support raw GMT/UTC offset strings in timezone validation

### LINKED ISSUE #5261: bug: Support raw GMT/UTC offsets in validation schema
- **Component**: Validation Service
- **Label**: `bug`, `validation`
- **Problem**: 
  The timezone validation schema accepts only named zones (e.g., `Europe/London`) and rejects raw offsets like `GMT+1` or `UTC-8`.
- **Solution**:
  1. Expand `isValidTimeZone` to validate raw GMT/UTC offset strings.
  2. Map valid offsets to safe timezone structures during validation.
- **Verification**:
  Pass `tz=GMT+5` and check that the schema returns `success: true`.

---

MasterJi27:
Fixes #5261

### DIFF
--- a/lib/validations.normalizeTimezone.test.ts
+++ b/lib/validations.normalizeTimezone.test.ts
@@ -1,0 +1,125 @@
+import { describe, it, expect } from 'vitest';
+import { normalizeTimezone } from './validations';
+import { streakParamsSchema } from './validations';
+
+describe('normalizeTimezone', () => {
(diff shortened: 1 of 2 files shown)


Aamod007 (approved):
Supporting raw GMT/UTC offset strings in timezone validation is a huge quality of life improvement for users not familiar with the tz database format. Handling the inverted sign convention (GMT+N -> Etc/GMT-N) inside normalizeTimezone is handled perfectly. Great test coverage as well. Approved!

PR #6121, opened 2026-06-20
fix(validations): coerce duplicate query parameter arrays into single values

### LINKED ISSUE #5267: bug: Coerce duplicate query parameter arrays into single values
- **Component**: Parameter Pollution
- **Label**: `security`, `bug`
- **Problem**: 
  Passing duplicate query keys (e.g. `?user=octocat&user=git`) results in Next.js parsing the values as an array, which crashes Zod schemas.
- **Solution**:
  1. In route validations, check if parameter values are arrays.
  2. If an array is passed, coerce it to its first string value before validation.
- **Verification**:
  Request `?user=octocat&user=git` and verify that the server validates and renders the badge for `octocat` successfully.

---

MasterJi27:
Fixes #5267

### DIFF
--- a/app/api/compare/route.ts
+++ b/app/api/compare/route.ts
@@ -1,7 +1,7 @@
 import { NextResponse } from 'next/server';
 import { getFullDashboardData } from '@/lib/github';
 import { getUserGitHubToken } from '@/lib/githubtoken';
-import { compareParamsSchema } from '@/lib/validations';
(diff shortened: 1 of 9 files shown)


Aamod007 (approved):
Handling duplicate query parameters securely is a classic edge case. Coercing query parameters using coerceQueryParams in app/api/compare/route.ts and others before passing them to the Zod schema ensures we don't hit unexpected validation failures when parameters are passed multiple times. Approved!

PR #6144, opened 2026-06-21
fix(cache): make DistributedCache.incr() fail closed on Redis errors

### LINKED ISSUE #6138: bug: DistributedCache.incr() silently falls back to per-instance local counter on Redis failure, defeating all rate limiting
### Description

`DistributedCache.incr()` in lib/cache.ts is the core primitive behind 
every rate limiter in the application (lib/rate-limit.ts calls 
`this.cache.incr()` for all per-IP rate limit checks).

When Redis is configured but a call fails (network error, timeout, 
Upstash outage), the catch block falls back to the LOCAL in-memory 
TTLCache as if it were the authoritative distributed counter:

```} catch (err) {
  logger.error('Cache INCR failed', {...});
  const current = (this.localCache.get(key) as unknown as number) || 0;
  const next = current + 1;
  ...
  return next;
}
```
Since serverless instances don't share memory, each instance maintains 
its own independent local counter starting from 0 whenever this fallback 
triggers. With N concurrent instances all hitting this fallback path 
during a Redis blip, an attacker effectively gets N times the intended 
rate limit completely defeating rate limiting across every protected 
endpoint in the application (/api/achievements, /api/compare, /api/og, 
/api/user-details, /api/ci-analytics, /api/notify, /api/webhook, etc.) 
for the duration of the Redis instability.

This also affects the non-Redis-error path's correctness expectations: 
once Redis recovers, the local fallback counters are completely out of 
sync with the real distributed count, with no reconciliation logic.

### Steps to Reproduce

1. Configure DistributedCache with valid Redis credentials
2. Simulate a Redis network failure (e.g. mock fetch to throw/timeout)
3. Call incr() from multiple separate DistributedCache instances 
   (simulating separate serverless instances) for the same key
4. Observe each instance returns its own independently-incrementing 
   count starting from 0/1, instead of a shared distributed count or 
   a fail-closed/fail-safe behavior

### Expected Behavior

On Redis failure, incr() should choose an explicit, documented failure 
mode rather than silently substituting a per-instance counter that 
masquerades as distributed state. Reasonable options:
1. Fail closed: throw, and let RateLimiter.check() treat the error as 
   "rate limit exceeded" (safer default degrades to blocking traffic 
   rather than allowing unlimited traffic during an outage)
2. Fail open with explicit logging/metrics distinguishing "distributed 
   count" from "degraded local-only count," so operators can detect 
   and respond to the condition, and so the local fallback counter is 
   reset/discarded once Redis recovers rather than silently persisting 
   skew

Either approach is preferable to the current behavior, which silently 
and indefinitely lets each instance operate on its own disconnected 
counter with no operator visibility into degraded rate-limiting state.

### Screenshots / Logs

_No response_

_No response_

### Environment

Other

taniy8:
## Description

Fixes #6138 

## Problem
`DistributedCache.incr()` in lib/cache.ts is the core primitive behind 
every rate limiter in the app (lib/rate-limit.ts calls cache.incr() for 
all per-IP rate limit checks). When the Redis call failed (network error, 
timeout, Upstash outage), the catch block silently fell back to the LOCAL 
in-memory TTLCache as if it were the authoritative distributed counter.

Serverless instances don't share memory, so each instance would maintain 
its own independent counter starting from 0 during any Redis blip. With 
N concurrent instances, an attacker effectively gets N times the intended 
rate limit defeating rate limiting across every protected endpoint 
(/api/achievements, /api/compare, /api/og, /api/user-details, /api/ci-analytics, 
/api/notify, /api/webhook, etc.) for the duration of the Redis instability.

## Fix
Changed incr() to fail closed on Redis errors instead of failing open. 
On failure, it now returns Number.MAX_SAFE_INTEGER, which 
RateLimiter.check() already correctly treats as "limit exceeded" 
(count > this.limit), no changes needed elsewhere. During a Redis 
outage, requests are now blocked rather than silently allowed through 
unlimited, which is the safer default for a security control.

Added a test simulating two separate DistributedCache instances 
(representing two serverless instances) both calling incr() on the 
same key while Redis is down, verifying both independently fail closed 
rather than each starting an unsynced local counter.

 
## Pillar

- [X] 🛠️ Other (Bug fix, refactoring, docs)

## Visual Preview
N/A

- [X] I have read the `CONTRIBUTING.md` file.
- [X] I have tested these changes locally (`localhost:3000/api/streak?user=YOUR_USERNAME`).
- [X] I have run `npm run format` and `npm run lint` locally and resolved all errors (CI will fail otherwise).
- [X] My commits follow the Conventional Commits format (e.g., `feat(themes): ...`, `fix(calculate): ...`).
- [ ] I have started the repo.
- [X] I have made sure that i have only one commit to merge in this PR.
- [X] (Recommended) I joined the CommitPulse Discord community for contributor discussions, mentorship, and faster PR support.


Aamod007 (approved):
Excellent catch on the fail-open vulnerability. Returning `Number.MAX_SAFE_INTEGER` inside `DistributedCache.incr()` is a very clean way to force the rate limiter to deny requests during a Redis outage, preventing the N-instance amplification attack vector. The `lib/cache.test.ts` update clearly proves the fail-closed behavior across mock instances. Approved!

PR #6145, opened 2026-06-21
feat(CherryBlossom): respect prefers-reduced-motion accessibility set…

### LINKED ISSUE #6142: feat: Respect prefers-reduced-motion for CherryBlossom background animation

The CherryBlossom component renders 25 petals that fall continuously across the entire viewport with `repeat: Infinity` in framer-motion — the animation never stops for as long as the page is open.

This is a problem for users with vestibular disorders or motion sensitivity who have set `prefers-reduced-motion: reduce` at the OS level. Right now the component completely ignores that preference and forces the animation on everyone regardless of their accessibility settings. This is exactly the kind of large-scale, perpetual background motion that WCAG 2.3.3 (Animation from Interactions) calls out as something users should be able to opt out of.

Use framer-motion's built-in `useReducedMotion()` hook inside CherryBlossom. If the user has `prefers-reduced-motion: reduce` set, the component should skip rendering the animated petals entirely (return null) instead of forcing the animation regardless of the user's OS-level preference.

This only requires:
1. Importing `useReducedMotion` from `framer-motion`
2. Calling it inside the component
3. Returning `null` early if `prefersReducedMotion` is true

No new dependencies needed — `useReducedMotion` already ships with framer-motion, which this component already depends on.

- Reducing petal count or animation speed instead of disabling entirely — rejected, since reduced motion should mean reduced motion, not just "slightly less" motion. A full stop is the safer, spec-aligned behavior.
- Manually wiring up a `window.matchMedia('(prefers-reduced-motion: reduce)')` listener — unnecessary, since framer-motion already exposes `useReducedMotion()` for this exact purpose and the project already depends on framer-motion.
- Leaving it as-is and relying on browser/OS-level animation throttling — not reliable, since `prefers-reduced-motion` is a user-facing accessibility signal that components are expected to respect directly, not something the browser silently handles for arbitrary CSS/JS animations.

aisshwaryaa8-collab:
…ting

## Description

CherryBlossom renders 25 falling petals on an infinite (repeat: Infinity) loop across the full viewport, with no way for the user to opt out. This adds a check using framer-motion's useReducedMotion() hook — if the user has prefers-reduced-motion: reduce set at the OS level, the component now returns null instead of forcing the animation on them. This aligns with WCAG 2.3.3 guidance on perpetual background motion.
Also updated the existing framer-motion mocks across 7 other CherryBlossom.*.test.tsx files to include useReducedMotion, since they previously only mocked motion.div and broke once the component started calling the new hook. Added 2 new tests to CherryBlossom.test.tsx covering both the reduced-motion and normal-motion paths.

Fixes #6142 

## Pillar

## Visual Preview
N/A


Aamod007 (approved):
Great focus on accessibility! Using `framer-motion`'s `useReducedMotion` hook in `components/CherryBlossom.tsx` ensures that we respect user OS preferences according to WCAG guidelines. Modifying the mocks across all the test files, like `components/CherryBlossom.empty-fallback.test.tsx`, is very thorough and prevents test regressions. Approved!

PR #6230, opened 2026-06-21
fix: prevent webhook memory exhaustion via Content-Length header bypass

### LINKED ISSUE #6135: Webhook endpoint allows memory exhaustion through Content-Length header bypass
## Description

The webhook endpoint in `app/api/webhook/route.ts` reads the entire request body into memory **before** validating its actual size. The first size check relies on the client-controlled `Content-Length` header, which is trivially forgeable.

```typescript
// Line 54-57 — Checks the DECLARED Content-Length (client-controlled)
const contentLength = Number(req.headers.get("content-length") || "0");
if (contentLength > MAX_PAYLOAD_SIZE) {
    return NextResponse.json({ error: "Payload too large" }, { status: 413 });
}

// Line 60-64 — Reads the ENTIRE body into memory
let bodyText: string;
try {
    bodyText = await req.text();  // <-- memory allocated here
} catch (error) {
    return NextResponse.json({ error: "Invalid payload" }, { status: 400 });
}

// Line 67-69 — Checks ACTUAL size AFTER full read (too late)
if (Buffer.byteLength(bodyText, "utf8") > MAX_PAYLOAD_SIZE) {
    return NextResponse.json({ error: "Payload too large" }, { status: 413 });
}
```

An attacker sends:
```
Content-Length: 100
Body: [1 GB of data]
```

Flow:
1. `contentLength` = 100 — passes the 1MB check
2. `req.text()` reads the **entire** 1GB body into a JavaScript string (allocating ~1-2GB memory)
3. The size check on line 67 detects the overflow, but **the memory is already consumed**

## Impact

- **On self-hosted deployments**: No built-in body size limit exists in Next.js App Router API routes (the Pages Router `bodyParser` config doesn't apply). An attacker can send arbitrarily large bodies, consuming all server memory and causing an OOM crash that kills the entire Node.js process and all in-flight requests.
- **On Vercel Pro/Enterprise**: Higher body limits than Hobby plan make this exploitable.
- **Single-request kill**: Only one request per minute is needed to exhaust gigabytes of memory — the 10/minute rate limit doesn't help because the damage occurs within a single request.
- **Webhook is publicly accessible**: It must receive GitHub webhook deliveries, so there's no IP allowlisting possible.

## Suggested Fix

Read the body with a streaming size limit instead of `req.text()`:

```typescript
import { ReadableStream } from "stream/web";

const MAX_PAYLOAD_SIZE = 1024 * 1024; // 1MB

// Use a TransformStream that enforces the size limit during reading
const reader = req.body!.getReader();
const chunks: Uint8Array[] = [];
let totalBytes = 0;

while (true) {
    const { done, value } = await reader.read();
    if (done) break;
    totalBytes += value.length;
    if (totalBytes > MAX_PAYLOAD_SIZE) {
        reader.cancel();
        return NextResponse.json({ error: "Payload too large" }, { status: 413 });
    }
    chunks.push(value);
}

const bodyText = new TextDecoder().decode(
    new Uint8Array(chunks.reduce((acc, chunk) => acc + chunk.length, 0))
);
```

Alternatively, configure a framework-level body size limit if supported by the Next.js version.

atul-upadhyay-7:
## Summary

Replaces the webhook endpoint's two-step payload size validation with a streaming `readBodyWithLimit()` function that enforces the 1MB size limit **during** chunked reading, preventing memory exhaustion attacks.

## Problem

The previous implementation checked the client-controlled `Content-Length` header before calling `req.text()`, which reads the **entire** body into memory as a JavaScript string. An attacker could:

1. Send `Content-Length: 100` (passes the 1MB check)
2. Send a 1GB actual body
3. `req.text()` allocates ~1-2GB of memory
4. The post-read size check detects the overflow, but the memory is already consumed

This allows a single request to crash the Node.js process via OOM.

## Solution

Added `readBodyWithLimit()` that reads the request body as a `ReadableStream` in chunks, tracking total bytes consumed. If the limit is exceeded at any point during reading, the stream is cancelled **immediately** — the full body is never buffered in memory.

```typescript
async function readBodyWithLimit(
  body: ReadableStream<Uint8Array> | null,
  maxBytes: number
): Promise<{ ok: true; body: string } | { ok: false; status: number; error: string }>
```

## Changes

- **`app/api/webhook/route.ts`** — Added `readBodyWithLimit()` function, replaced `Content-Length` check + `req.text()` with streaming read
- **`app/api/webhook/route.test.ts`** — Added regression test: "returns 413 when actual body exceeds 1MB despite small Content-Length"

## Testing

- All 9 webhook tests pass (including the new vulnerability regression test)
- Lint: 0 errors
- Typecheck: no new errors

Fixes #6135

Human Coded

### DIFF
--- a/app/api/webhook/route.test.ts
+++ b/app/api/webhook/route.test.ts
@@ -116,6 +116,22 @@ describe('POST /api/webhook', () => {
     expect(data.error).toBe('Payload too large');
   });
 
+  it('returns 413 when actual body exceeds 1MB despite small Content-Length', async () => {
+    process.env.GITHUB_WEBHOOK_SECRET = 'secret_key';
+    const largePayload = 'x'.repeat(1024 * 1024 + 1);
+    const req = makeRequest(
+      {
+        'content-length': '100',
+        'x-hub-signature-256': 'sha256=somesignature',
+      },
+      largePayload
+    );
+    const res = await POST(req);
+    expect(res.status).toBe(413);
+    const data = await res.json();
+    expect(data.error).toBe('Payload too large');
+  });
+
   it('returns 429 when rate limit is exceeded', async () => {
     process.env.GITHUB_WEBHOOK_SECRET = 'secret_key';
     const secret = 'secret_key';
--- a/app/api/webhook/route.ts
+++ b/app/api/webhook/route.ts
@@ -6,6 +6,7 @@ import { logger } from '@/lib/logger';
 const MAX_PAYLOAD_SIZE = 1024 * 1024; // 1MB
 const SIGNATURE_PREFIX = 'sha256=';
 const SHA256_HEX_LENGTH = 64;
+const READ_CHUNK_SIZE = 64 * 1024; // 64KB chunks
 
 function getWebhookSecret(): string | null {
   const secret = process.env.GITHUB_WEBHOOK_SECRET?.trim();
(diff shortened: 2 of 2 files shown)


Aamod007 (approved):
Excellent fix. Replacing req.text() with the streaming readBodyWithLimit() chunk reader in app/api/webhook/route.ts effectively neutralizes the out-of-memory vulnerability caused by forged Content-Length headers. Adding the regression test in app/api/webhook/route.test.ts to simulate the mismatch ensures this doesn't regress. Approved!

Issue #6242, opened 2026-06-21
bug: wrapped route calculates default year using server timezone instead of user's tz parameter


nishupr:
Problem
In app/api/wrapped/route.ts, the default year (when customYear is not provided) is calculated using the server's local timezone:

const year = customYear || new Date().getFullYear().toString();

However, the route already accepts and uses a `tz` parameter for fetching timezone-aware wrapped stats:

const wrappedStats = tz
  ? await getWrappedData(user, year, { bypassCache: shouldBypassCache }, tz)
  : await getWrappedData(user, year, { bypassCache: shouldBypassCache });

This creates an inconsistency: a user near a year boundary (e.g. New Year's Eve/Day) in a different timezone than the server could get the wrong default year. For example, if the server is on UTC and a user is in UTC+13 (New Zealand), the server might still report the previous year while the user has already entered the new year locally — or vice versa.

This mirrors a pattern already correctly handled elsewhere in the codebase, e.g. app/api/streak/route.ts:

new Intl.DateTimeFormat('en-CA', { timeZone: timezone, year: 'numeric' }).format(new Date())

Root cause
The year calculation in wrapped/route.ts was written using plain new Date().getFullYear() without considering the tz parameter that the route already supports for other purposes.

Proposed fix
Use the tz parameter (when provided) to compute the default year, consistent with the timezone-aware pattern in streak/route.ts:

const year = customYear || (tz
  ? new Intl.DateTimeFormat('en-CA', { timeZone: tz, year: 'numeric' }).format(new Date())
  : new Date().getFullYear().toString());

PR #6246, opened 2026-06-21
fix(validations): use UTC year for streak/wrapped year validation upper bound

### LINKED ISSUE #6244: bug: year validation in validations.ts uses server timezone instead of UTC for currentYear upper bound
Problem
In lib/validations.ts, both streakParamsSchema and wrappedParamsSchema validate the year query parameter using:

const currentYear = new Date().getFullYear();
return /^\d{4}$/.test(val) && yearNum >= 2008 && yearNum <= currentYear;

new Date().getFullYear() returns the year in the server's local timezone, not UTC or the user's timezone. Near a year boundary (e.g. Dec 31 / Jan 1), this can cause a valid year value submitted by a user to be incorrectly rejected (or an invalid one to be incorrectly accepted), depending on the server's local time relative to UTC.

This is the same class of inconsistency already addressed in app/api/wrapped/route.ts (see PR #6242 / issue #6242), where the default year calculation was changed to use the tz parameter via Intl.DateTimeFormat instead of the server's local time.

Root cause
The currentYear upper bound check was written using new Date().getFullYear() (server-local time) instead of a UTC-based or timezone-aware calculation.

Proposed fix
Replace the local-time calculation with a UTC-based one for consistency and predictability:

const currentYear = new Date().getUTCFullYear();

This avoids server-timezone-dependent validation behavior. (Note: the refine() callback only receives the field value, not other parsed fields like tz, so full per-user timezone awareness would require superRefine — but switching to UTC alone removes the server-local-time dependency, which is the core bug.)

nishupr:
## Description
Fixes #6244

In `lib/validations.ts`, both `streakParamsSchema` and `wrappedParamsSchema` validate the `year` query parameter's upper bound using:

```ts
const currentYear = new Date().getFullYear();
return /^\d{4}$/.test(val) && yearNum >= 2008 && yearNum <= currentYear;
```

**Problem with the old approach:**
- `new Date().getFullYear()` returns the year in the **server's local timezone**, not UTC
- Near a year boundary (e.g. Dec 31 / Jan 1), this can cause inconsistent validation behavior depending on the server's local time offset from UTC
- This is the same class of inconsistency already addressed in `app/api/wrapped/route.ts` (#6242), where the default year calculation was changed to avoid server-local-time dependency

**What this PR does:**
- Replaces `new Date().getFullYear()` with `new Date().getUTCFullYear()` in both `streakParamsSchema` and `wrappedParamsSchema` validation refinements
- Updates the corresponding test file `lib/validations.year.test.ts` to compute `CURRENT_YEAR` via `getUTCFullYear()` for consistency, ensuring tests remain accurate regardless of the test runner's local timezone
- No change to the validation logic itself — only removes the server-timezone dependency

**Note:** The `refine()` callback only receives the field value, not other parsed fields like `tz`, so full per-user timezone-aware validation would require `superRefine`. Switching to UTC removes the core server-local-time dependency, which was the primary issue.

**After:**
```ts
const currentYear = new Date().getUTCFullYear();
return /^\d{4}$/.test(val) && yearNum >= 2008 && yearNum <= currentYear;
```

## Pillar
- [x] 🕐 Pillar 3 — Timezone Logic Optimization
- [ ] 🛠️ Other (Bug fix, refactoring, docs)

## Visual Preview
No visual changes — this is a validation correctness fix.

### DIFF
--- a/lib/validations.ts
+++ b/lib/validations.ts
@@ -321,7 +321,7 @@ const baseStreakParamsSchema = z.object({
       (val) => {
         if (!val) return true;
         const yearNum = parseInt(val, 10);
-        const currentYear = new Date().getFullYear();
+        const currentYear = new Date().getUTCFullYear();
         return /^\d{4}$/.test(val) && yearNum >= 2008 && yearNum <= currentYear;
       },
       {
@@ -672,7 +672,7 @@ export const wrappedParamsSchema = z.object({
       (val) => {
         if (!val) return true;
         const yearNum = parseInt(val, 10);
-        const currentYear = new Date().getFullYear();
+        const currentYear = new Date().getUTCFullYear();
         return /^\d{4}$/.test(val) && yearNum >= 2008 && yearNum <= currentYear;
       },
       {
--- a/lib/validations.year.test.ts
+++ b/lib/validations.year.test.ts
@@ -1,7 +1,7 @@
 import { describe, it, expect } from 'vitest';
 import { streakParamsSchema, wrappedParamsSchema } from './validations';
 
-const CURRENT_YEAR = new Date().getFullYear();
+const CURRENT_YEAR = new Date().getUTCFullYear();
 const EXPECTED_ERROR = 'GitHub was founded in 2008. Please provide a year of 2008 or later.';
 
 function getYearError(result: {


Aamod007 (approved):
Using getUTCFullYear() instead of getFullYear() in lib/validations.ts is a critical fix to ensure we don't have timezone-dependent validation failures around the New Year boundary. The update to lib/validations.year.test.ts to reflect this change is also perfectly executed. Approved!

PR #6259, opened 2026-06-21
fix(validation): validate managementToken format using regex

### LINKED ISSUE #6192: fix(validation): managementToken format not validated - accepts any string (MEDIUM)
## Description

In `lib/validations.ts` at line 734, the `managementToken` field in the Zod validation schema accepts **any string** between 16 and 256 characters:

```typescript
managementToken: z.string().min(16).max(256)
```

Management tokens follow a specific format — they start with a `cpn_` prefix and are base64url-encoded. However, the validation does not enforce this format. When an invalid token is used, it fails later with a cryptic error that is hard to debug, typically surfacing as a generic authentication failure.

### Impact

**MEDIUM** — Poor user experience and debuggability:

1. Users who accidentally paste a truncated or malformed token get a confusing error message
2. Developers debugging integration issues cannot tell if the problem is the token format or an actual authentication failure
3. Support requests increase as users struggle to understand why their token doesn't work

### Location

`lib/validations.ts:734`

### Suggested Fix

Add a regex refinement to validate the token format upfront:

```typescript
managementToken: z.string()
    .min(16)
    .max(256)
    .regex(/^cpn_[A-Za-z0-9_-]+$/, 'Invalid management token format — must start with "cpn_" and be base64url-encoded')
```

This gives users immediate, clear feedback when they provide a malformed token.

Abfa41:
## Description

Fixes #6192

This PR adds format validation for the `managementToken` field to ensure malformed tokens are rejected during schema validation instead of failing later during authentication.

## Changes made

* Added regex validation for `managementToken`.
* Enforced the required `cpn_` prefix.
* Restricted token characters to base64url-safe characters (`A-Z`, `a-z`, `0-9`, `_`, `-`).
* Preserved existing length constraints and optional behavior.
* Added a clear validation error message for malformed tokens.
* Updated notification route test fixtures to use management tokens that match the enforced `cpn_` token format.

## Benefits

* Provides immediate feedback for invalid tokens.
* Improves debuggability for developers.
* Reduces confusion caused by generic authentication failures.
* Prevents malformed values from progressing further into the application.

## Validation Rules

Accepted format:

```text
cpn_xxxxxxxxxxxxxxxxx
```

Rejected examples:

```text
abc_token
cpn token
cpn@token
```

### DIFF
--- a/app/api/notify/route.test.ts
+++ b/app/api/notify/route.test.ts
@@ -301,7 +301,7 @@ describe('POST /api/notify', () => {
   });
 
   it('allows updates with a valid notification management token', async () => {
-    const managementToken = 'valid-management-token';
+    const managementToken = 'cpn_valid-management-token';
     vi.mocked(Notification.findOne).mockResolvedValue({
       username: 'tokenuser',
       email: 'old@example.com',
--- a/lib/validations.ts
+++ b/lib/validations.ts
@@ -815,7 +815,16 @@ export const notifyPostSchema = z.object({
       notifyOnStreak: true,
       notifyOnMilestone: true,
     }),
-  managementToken: z.string().trim().min(16).max(256).optional(),
+  managementToken: z
+    .string()
+    .trim()
+    .min(16)
+    .max(256)
+    .regex(
+      /^cpn_[A-Za-z0-9_-]+$/,
+      'Invalid management token format — must start with "cpn_" and be base64url-encoded'
+    )
+    .optional(),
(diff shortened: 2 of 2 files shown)


Aamod007 (approved):
GSSoC 2025 — @Aamod007

Zod regex validation for managementToken (cpn_ prefix + base64url chars). Small but real security hardening. level:beginner / quality:clean / type:bug / type:security

PR #6306, opened 2026-06-22
fix: remove token from GIT_ASKPASS script body (#6287)

### LINKED ISSUE #6287: GitHub OAuth token persisted to /tmp in plaintext shell script
## Impact: GitHub OAuth token exfiltration

In `app/api/architecture/route.ts`, the user's decrypted GitHub OAuth token is written to a shell script on disk:

```javascript
fs.writeFileSync(askpassScript, `#!/bin/sh\necho "${token}"`)
```

This script is created in `/tmp` with mode `0o700`, but:
- The token value is embedded directly in the script body — no escaping, no environment variable
- If the Node.js process is killed (OOM, SIGKILL, cold start timeout), the `finally` block never runs
- The script remains readable in `/tmp` until the OS purges it (could be hours/days)
- On shared hosting or container environments, other processes may read `/tmp`

**This is a token exfiltration path.** A crashed process = a leaked GitHub token with full repo access, push rights, and OAuth scope.

**File:** `app/api/architecture/route.ts` lines 295-310

**Fix:**
```javascript
// Instead of writing token to disk:
const child = execFile('git', ['clone', url, dir], {
  env: { ...process.env, GIT_ASKPASS: 'echo', GIT_TERMINAL_PROMPT: '0' },
  // Or better: use GIT_ASKPASS with a Node script that reads from env
});
```
Never persist secrets to the filesystem. Use `child_process.env` to pass tokens in-memory only.

atul-upadhyay-7:
## Summary

This PR fixes Issue #6287 by removing the GitHub OAuth token from the GIT_ASKPASS script body.

## Problem

The user's decrypted GitHub OAuth token was being written directly to a shell script on disk:

```javascript
fs.writeFileSync(askpassScript, `#!/bin/sh\necho "${token}"`, { mode: 0o700 });
```

If the Node.js process was killed (OOM, SIGKILL, cold start timeout), the `finally` block never runs, and the script remains readable in `/tmp` until the OS purges it. This is a token exfiltration path.

## Solution

The token is now passed via the `GIT_TOKEN` environment variable instead of being embedded in the script body. The script reads from the environment variable:

```javascript
const askpassScript = path.join(tempDir, '.git-askpass.sh');
fs.writeFileSync(
  askpassScript,
  '#!/bin/sh\nif [ -n "$GIT_TOKEN" ]; then\necho "$GIT_TOKEN"\nelse\nexit 1\nfi',
  { mode: 0o700 }
);
env.GIT_ASKPASS = askpassScript;
env.GIT_TOKEN = token;
```

This ensures:
- The token is never written to disk in plaintext
- The token is only in memory during the git clone operation
- If the process crashes, the token is not persisted

Fixes #6287

### DIFF
--- a/app/api/architecture/route.ts
+++ b/app/api/architecture/route.ts
@@ -296,11 +296,17 @@ export async function POST(req: NextRequest) {
     try {
       const env = { ...process.env } as NodeJS.ProcessEnv;
       if (token) {
-        // GIT_ASKPASS points to a script that echoes the token when git asks
-        // for credentials, keeping the token out of process arguments.
+        // GIT_ASKPASS points to a script that reads the token from an environment
+        // variable instead of embedding it in the script body. This prevents the
+        // token from persisting on disk if the process crashes.
         const askpassScript = path.join(tempDir, '.git-askpass.sh');
-        fs.writeFileSync(askpassScript, `#!/bin/sh\necho "${token}"`, { mode: 0o700 });
+        fs.writeFileSync(
+          askpassScript,
+          '#!/bin/sh\nif [ -n "$GIT_TOKEN" ]; then\necho "$GIT_TOKEN"\nelse\nexit 1\nfi',
+          { mode: 0o700 }
+        );
         env.GIT_ASKPASS = askpassScript;
+        env.GIT_TOKEN = token;
         env.GIT_TERMINAL_PROMPT = '0';
       }
       await execFilePromise('git', ['clone', '--depth', '1', '--', cloneUrl, tempDir], { env });


Aamod007 (approved):
## GSSoC 2025 Mentor Review — @Aamod007

**Labels:** level:intermediate, quality:clean, type:security, type:bug, mentor:Aamod007

**Difficulty (intermediate):** Addresses a token-exfiltration vulnerability via disk persistence in pp/api/architecture/route.ts:302-308. Requires understanding of process lifecycle, shell script execution, and environment variable scoping. Security-critical but single file.

**Quality (clean):** Minimal and correct fix — the GIT_ASKPASS script body uses \ from environment with a fallback exit 1, and the token is scoped per execFilePromise call via env object.

**Type (security + bug):** Removes token-on-disk attack surface — a crash (OOM/SIGKILL) would leave the token in /tmp scripts. Fixes #6287.

LGTM — moving the token to env.GIT_TOKEN is exactly the right pattern here. Thanks for catching this!

PR #6338, opened 2026-06-22
fix(ci-analytics): guard successRate against divide-by-zero and exclude zero-duration runs from avg

### LINKED ISSUE #6337: bug: /api/ci-analytics successRate returns NaN (null) when no run has a success or failure conclusion
## Description

`/api/ci-analytics` returns a broken `successRate` (and a deflated `avgBuildDuration`) for repositories whose recent GitHub Actions runs were not all decided as success/failure.

In `services/github/ci-analytics.ts`, `processRuns` computes:

```ts
const successRate =
  totalRuns > 0 ? Math.round((successfulRuns / (successfulRuns + failedRuns)) * 100) : 0;
```

The guard checks `totalRuns > 0`, but the divisor is `successfulRuns + failedRuns`. If a repo's fetched runs are all `cancelled`, `skipped`, `queued`, or `in_progress` (so `successfulRuns = 0` and `failedRuns = 0`) while `totalRuns > 0`, the expression evaluates to `Math.round(0 / 0)` = `Math.round(NaN)` = `NaN`, which `JSON.stringify` serializes to `null`. So `successRate` comes back as `null` instead of a number.

Separately, `avgBuildDuration` divides the total duration by `completedRuns.length`, where `completedRuns` is filtered only by `r.conclusion !== null`. Runs whose duration could not be computed fall back to `0` (see `calculateDurationFallback`), and those zero-duration runs stay in the denominator, pulling the average down. This is inconsistent with `buildInsights` (and the spirit of `buildRepoHealth`), which already require `r.duration > 0`.

## Steps to Reproduce

1. `GET /api/ci-analytics?username=<user>` where the 5 most-recently-pushed repos have GitHub Actions runs that are only `cancelled` / `in_progress` (no `success` or `failure`).
2. Inspect the JSON response: `successRate` is `null` instead of a number.
3. Static trace: in `services/github/ci-analytics.ts`, read the `successRate` line in `processRuns`; with `successfulRuns = 0` and `failedRuns = 0` but `totalRuns > 0`, the divisor is `0`, so the result is `Math.round(NaN)` = `NaN` (serialized as `null`).

## Expected Behavior

`successRate` is always a number. When there are no decided runs (no success and no failure), it should be `0` (matching the existing fallback), never `NaN`/`null`. `avgBuildDuration` should be averaged only over runs that have a real (`> 0`) duration, consistent with `buildInsights`.

## Screenshots / Logs

Static reproduction from the cited code is definitive; a response capture for an all-cancelled repo would show `"successRate": null`.

## Environment

Other

Anexus5919:
## Description

Fixes #6337

`processRuns` in `services/github/ci-analytics.ts` computed `successRate` by guarding on `totalRuns > 0` while dividing by `successfulRuns + failedRuns`. For a repo whose fetched runs are all `cancelled` / `skipped` / `queued` / `in_progress` (so `successfulRuns = 0` and `failedRuns = 0`) but `totalRuns > 0`, this is `Math.round(0 / 0)` = `NaN`, which serializes to `null` in the API response.

This change guards on the actual denominator:

```ts
const decidedRuns = successfulRuns + failedRuns;
const successRate = decidedRuns > 0 ? Math.round((successfulRuns / decidedRuns) * 100) : 0;
```

So with no decided runs the rate is `0` (the existing fallback value), never `NaN`/`null`. Repos with at least one success/failure are unaffected (the rate is still success over decided runs).

It also fixes `avgBuildDuration`, which divided by `completedRuns.length` where `completedRuns` only required `conclusion !== null`. Runs whose duration could not be computed fall back to `0` and were inflating the denominator. The filter now also requires `r.duration > 0`, matching `buildInsights`:

```ts
const completedRuns = runs.filter((r) => r.conclusion !== null && r.duration > 0);
```

`processRuns` is now exported so the aggregation can be unit-tested directly.

## Pillar

## Visual Preview

No UI change. The `/api/ci-analytics` JSON now returns a numeric `successRate` (0 when there are no decided runs) instead of `null`, and `avgBuildDuration` excludes zero-duration runs.

- [x] I have tested these changes locally (new `services/github/ci-analytics.test.ts`, 3 tests, plus the existing route test pass; `tsc --noEmit` clean).

### DIFF
--- a/services/github/ci-analytics.test.ts
+++ b/services/github/ci-analytics.test.ts
@@ -1,0 +1,61 @@
+import { describe, it, expect } from 'vitest';
+import { processRuns } from './ci-analytics';
+import type { CIWorkflowRun } from '@/types/ci-analytics';
+
+function makeRun(overrides: Partial<CIWorkflowRun>): CIWorkflowRun {
+  return {
+    id: 1,
+    name: 'CI',
+    repository: 'owner/repo',
+    branch: 'main',
+    status: 'completed',
+    conclusion: 'success',
+    duration: 60,
+    triggerEvent: 'push',
+    startedAt: '2026-01-01T00:00:00Z',
+    finishedAt: '2026-01-01T00:01:00Z',
+    url: 'https://example.test',
+    ...overrides,
+  };
+}
+
+describe('processRuns', () => {
+  it('returns a 0 success rate (never NaN) when no run has a success or failure conclusion', () => {
+    const runs = [
+      makeRun({ conclusion: 'cancelled' }),
+      makeRun({ conclusion: null, status: 'in_progress' }),
+      makeRun({ conclusion: 'skipped' }),
+    ];
+
+    const result = processRuns(runs);
+
+    expect(result.totalRuns).toBe(3);
+    expect(result.successRate).toBe(0);
+    expect(Number.isNaN(result.successRate)).toBe(false);
+  });
+
+  it('computes the success rate over decided runs only (success + failure)', () => {
+    const runs = [
+      makeRun({ conclusion: 'success' }),
+      makeRun({ conclusion: 'success' }),
+      makeRun({ conclusion: 'success' }),
+      makeRun({ conclusion: 'failure' }),
+      makeRun({ conclusion: 'cancelled' }),
+      makeRun({ conclusion: null, status: 'queued' }),
+    ];
+
+    // 3 success out of 4 decided runs = 75%; cancelled and queued runs are excluded.
+    expect(processRuns(runs).successRate).toBe(75);
+  });
+
+  it('excludes zero-duration completed runs from the average build duration', () => {
+    const runs = [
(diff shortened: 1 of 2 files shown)


Aamod007 (approved):
**Labels:** level:intermediate, quality:clean, type:bug, mentor:Aamod007

**Justification — difficulty (intermediate):** Requires reasoning about the NaN edge case (successRate = Math.round(0/0) when all runs are cancelled/skipped), understanding the denominator distinction (decidedRuns vs totalRuns), and the zero-duration inflation bug in avgBuildDuration.

**Justification — quality (clean):** 3 tests covering NaN prevention, decided-runs-only denominator, and zero-duration filter. Exports processRuns for direct unit testing without HTTP.

**Justification — type (bug):** successRate was NaN→null in API response for repos with only cancelled/skipped runs. avgBuildDuration was inflated by zero-duration runs.

Clean edge-case fix, merging.

PR #6342, opened 2026-06-22
fix(burnout-analyzer): validate exactly owner/repo and URL-encode the request params

### LINKED ISSUE #6341: bug: /burnout-analyzer accepts malformed repo paths (3+ segments analyzed as owner/repo; input not URL-encoded)
## Description

The Burnout Analyzer page (`app/burnout-analyzer/page.tsx`) mishandles the "owner/repo" text input in two ways.

1. Extra-segment drop. The only validation is `repoPath.includes('/')`, followed by `const [owner, repo] = repoPath.trim().split('/')`. A path with more than two segments (for example a pasted URL fragment like `facebook/react/tree/main`) passes the check, and the destructuring keeps only the first two segments, so the app silently analyzes `facebook/react` and discards `tree/main`. The user gets a result for a repo they did not ask for, with no error.

2. Missing URL-encoding. `owner` and `repo` are interpolated directly into the request URL: `fetch(`/api/repo-burnout?owner=${owner}&repo=${repo}`)` (and the same in the refresh handler). If the typed value contains a character such as `&`, `#`, or a space, the query string is corrupted. For example typing `foo/bar&x=1` produces `...?owner=foo&repo=bar&x=1`, so the server receives `repo=bar` (a different, valid repo) plus a spurious `x=1` param, instead of validating the value the user actually typed. The `/api/repo-burnout` route validates `owner`/`repo` with strict regexes, so this is a client-side correctness/UX issue, not a server-side injection.

## Steps to Reproduce

1. Open `/burnout-analyzer`.
2. Enter `facebook/react/tree/main` and click Analyze. Observe it analyzes `facebook/react` (the `tree/main` part is silently dropped) instead of rejecting the malformed path.
3. Enter `foo/bar&x=1` and click Analyze. The request URL becomes `/api/repo-burnout?owner=foo&repo=bar&x=1`, so the server sees `repo=bar` rather than the typed `bar&x=1`.

## Expected Behavior

- A path that is not exactly two non-empty segments is rejected with the existing "valid repository path" error.
- `owner` and `repo` are URL-encoded so the value the user typed is exactly what is sent and validated.

## Screenshots / Logs

Reproducible from the steps above; the cited lines make it deterministic.

## Environment

Other

Anexus5919:
## Description

Fixes #6341

`app/burnout-analyzer/page.tsx` mishandled the owner/repo text input:

- The only validation was `repoPath.includes('/')`, then `const [owner, repo] = ...split('/')`, so a path with extra segments (for example `facebook/react/tree/main`) passed and was silently analyzed as `facebook/react` (the rest dropped).
- `owner`/`repo` were interpolated into the request URL without encoding, so a typed value containing `&`, `#`, or a space corrupted the query string (for example `foo/bar&x=1` became `...?owner=foo&repo=bar&x=1`, so the server received a different repo plus a spurious param).

This change:

- Validates exactly two non-empty segments and rejects anything else with the existing error message.
- URL-encodes `owner` and `repo` with `encodeURIComponent` in both the analyze and refresh requests, so the value the user typed is exactly what is sent (and validated by the route).

Scope note: I also looked at the request-race and "disable the button while loading" angles and found them already mitigated by the existing UI. When `isLoading` is true the search form (with its Analyze button and the suggestion buttons) unmounts and is replaced by the loading skeleton, so a second concurrent analyze cannot be triggered from this page; and the refresh button is already `disabled={isRefreshing}` in `RepoHeader`. So I did not add an AbortController or further button-disabling, to keep the change minimal.

Added `app/burnout-analyzer/page.test.tsx` covering the rejection of a 3-segment path (no API call is made) and the URL-encoding of special characters; both tests fail against the old code.

## Pillar

## Visual Preview

No layout change. Malformed paths now show the existing validation error instead of silently analyzing a different repo, and special characters in the input are encoded into the request.

- [x] I have tested these changes locally (new `app/burnout-analyzer/page.test.tsx`, 3 tests, pass; `tsc --noEmit` clean).

### DIFF
--- a/app/burnout-analyzer/page.test.tsx
+++ b/app/burnout-analyzer/page.test.tsx
@@ -1,0 +1,61 @@
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+import { render, screen, fireEvent, waitFor } from '@testing-library/react';
+import BurnoutAnalyzerPage from './page';
+
+describe('BurnoutAnalyzerPage repository input handling', () => {
+  beforeEach(() => {
+    vi.clearAllMocks();
+  });
+
+  it('rejects a path that is not exactly owner/repo and does not call the API', async () => {
+    const fetchMock = vi.fn();
+    vi.stubGlobal('fetch', fetchMock);
+
+    render(<BurnoutAnalyzerPage />);
+    fireEvent.change(screen.getByPlaceholderText(/facebook\/react/i), {
+      target: { value: 'facebook/react/tree/main' },
+    });
+    fireEvent.click(screen.getByRole('button', { name: /analyze/i }));
+
+    expect(await screen.findByText(/valid repository path/i)).toBeInTheDocument();
+    expect(fetchMock).not.toHaveBeenCalled();
+  });
+
+  it('URL-encodes owner and repo so the typed input is what gets validated, not a different repo', async () => {
+    const fetchMock = vi.fn().mockResolvedValue({
+      ok: false,
+      json: async () => ({ error: 'Invalid repo name format' }),
+    });
+    vi.stubGlobal('fetch', fetchMock);
+
+    render(<BurnoutAnalyzerPage />);
+    fireEvent.change(screen.getByPlaceholderText(/facebook\/react/i), {
+      target: { value: 'foo/bar&x=1' },
+    });
+    fireEvent.click(screen.getByRole('button', { name: /analyze/i }));
+
+    await waitFor(() => expect(fetchMock).toHaveBeenCalled());
+    const calledUrl = fetchMock.mock.calls[0][0] as string;
+    expect(calledUrl).toContain(`repo=${encodeURIComponent('bar&x=1')}`);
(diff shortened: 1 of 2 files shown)


Aamod007 (approved):
**Labels:** level:intermediate, quality:clean, type:bug, type:security, mentor:Aamod007

**Justification — difficulty (intermediate):** Covers input validation (exactly 2 segments), URL encoding with encodeURIComponent, and 3 test cases. Requires understanding the injection vector and implementing proper sanitization.

**Justification — quality (clean):** Thorough description documenting both bugs (silent truncation, parameter injection), tests that fail against old code, security-conscious encoding on both analyze and refresh endpoints.

**Justification — type (bug, security):** Bug — multi-segment paths like 'facebook/react/tree/main' were silently analyzed as 'facebook/react'. Security — encodeURIComponent prevents query parameter injection via typed input containing &, #, etc.

Clean fix, merging.

PR #6386, opened 2026-06-23
fix: close TOCTOU race in refresh policy by atomically checking and recording

### LINKED ISSUE #6381: Race condition in refresh policy allows concurrent cache bypasses, exhausting GitHub API quota
## Summary
The `refreshPolicy.isRefreshAllowed()` check and `refreshPolicy.recordRefresh()` call in `app/api/stats/route.ts` are separated by the entire data fetch operation. This creates a check-then-act race condition where multiple concurrent requests for the same user can all bypass the cache simultaneously.

## Affected File
- `app/api/stats/route.ts` (lines 90-136)

## Steps to Reproduce
1. Send 5+ concurrent requests to `/api/stats?user=someuser&refresh=true`
2. All requests pass the `isRefreshAllowed()` check before any calls `recordRefresh()`
3. All requests bypass the cache and hit the GitHub API

## Root Cause
The cooldown check and recording are not atomic:

```typescript
// Line 114 — check
if (!refreshPolicy.isRefreshAllowed(user)) {
  shouldBypassCache = false;
}

// ... full data fetch happens here ...

// Line 135 — act
refreshPolicy.recordRefresh(user);
```

Concurrent requests interleave between the check and the record, defeating the 30-second cooldown.

## Impact
- Excessive GitHub API calls per user during cooldown windows
- Potential rate limit exhaustion for the shared token pool
- Degraded service for all users during traffic spikes

## Suggested Fix
Use an atomic operation — record the refresh **before** the data fetch, or use a mutex/lock:

```typescript
if (isRefreshRequested) {
  if (!refreshPolicy.tryAcquire(user)) {  // atomic check-and-set
    shouldBypassCache = false;
  }
}
// ... data fetch ...
// No need for separate recordRefresh() call
```

anshika1179:
## Summary

Fixes a TOCTOU (Time-of-Check-to-Time-of-Use) race condition in the refresh policy that allowed concurrent requests to bypass the cache simultaneously, exhausting GitHub API quota.

## Changes

- **`services/github/refresh-policy.ts`**: Added `tryAcquire()` method that atomically checks whether a refresh is allowed and records it in a single operation, closing the race window.
- **`app/api/stats/route.ts`**: Replaced separate `isRefreshAllowed()` + `recordRefresh()` calls with the new atomic `tryAcquire()` method.

## How it fixes the race

**Before:** Concurrent requests could all pass `isRefreshAllowed()` before any called `recordRefresh()`, defeating the 30-second cooldown.

**After:** `tryAcquire()` checks and records in one step. Only the first concurrent request succeeds; subsequent requests within the cooldown window are rejected.

## Closes

Closes #6381

### DIFF
--- a/app/api/stats/route.ts
+++ b/app/api/stats/route.ts
@@ -111,7 +111,7 @@ export async function GET(request: Request) {
 
   let shouldBypassCache = isRefreshRequested;
   if (isRefreshRequested) {
-    if (!refreshPolicy.isRefreshAllowed(user)) {
+    if (!refreshPolicy.tryAcquire(user)) {
       logSecurityEvent('STATS_REFRESH_COOLDOWN_VIOLATION', {
         user,
         ip,
@@ -131,9 +131,6 @@ export async function GET(request: Request) {
       bypassCache: shouldBypassCache,
       token: userToken,
     });
-    if (shouldBypassCache) {
-      refreshPolicy.recordRefresh(user);
-    }
 
     const calendar = userData.calendar;
     const stats = calculateStreak(calendar, timezone);
--- a/services/github/refresh-policy.ts
+++ b/services/github/refresh-policy.ts
@@ -72,6 +72,25 @@ export class RefreshPolicy {
   }
 
   /**
+   * Atomically checks whether a refresh is allowed and, if so, records it.
+   *
+   * This eliminates the TOCTOU race condition between `isRefreshAllowed()`
+   * and `recordRefresh()` where concurrent requests could all pass the
+   * cooldown check before any of them recorded the refresh.
+   *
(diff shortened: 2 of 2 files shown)


Aamod007 (changes requested):
Hello! Great job closing the TOCTOU race condition. By replacing the separate check and record calls with an atomic tryAcquire() in app/api/stats/route.ts, you've ensured concurrent requests cannot bypass the 30-second cooldown window. This prevents GitHub API quota exhaustion effectively. However, the Vercel build is failing right now, so this PR is blocked. Let's get the CI passing!

Aamod007 (approved):
The Vercel CI failure is a known global issue and can be ignored. Since there are no merge conflicts with main, I am officially approving this PR! Great work! 🚀

PR #6411, opened 2026-06-23
fix(cache): resolve localLocks memory leak in DistributedCache 

### LINKED ISSUE #6177: fix(cache): localLocks Map unbounded growth - memory leak in DistributedCache (CRITICAL)
## Description

In `lib/cache.ts` at line 321, the `DistributedCache` class maintains a `localLocks` map with **no maximum size or eviction policy**:

```typescript
private localLocks = new Map<string, Promise<T>>();
```

Entries are added on every `getOrSet` call and removed via `.finally()` in the cleanup handler (line 773-775). However, if a `loadFn` promise **never resolves or rejects** (e.g., a hanging fetch, an infinite loop, or a promise that is simply abandoned), the corresponding entry in `localLocks` is never cleaned up.

### Impact

**CRITICAL** — Memory leak under sustained production load. Over time:

1. Unique keys accumulate in the map without bound
2. Each entry holds a reference to a `Promise<T>` that may keep other objects alive
3. Garbage collection cannot reclaim these entries
4. Memory usage grows linearly with unique cache key count
5. Eventually leads to OOM (Out of Memory) crashes in long-running server processes

### Location

`lib/cache.ts:321`

### Suggested Fix

Implement one or both of:

1. **LRU eviction**: Wrap the map with a maximum size (e.g., 1000 entries) using a library like `lru-cache` or a manual `Map` with an LRU tracking list
2. **TTL-based cleanup**: Add a timeout to each lock entry that auto-removes it after a reasonable duration (e.g., 30 seconds)

```typescript
private localLocks = new Map<string, { promise: Promise<T>; timestamp: number }>();
// Cleanup stale entries on each access
```

Mohammedsami001:
## Description

Fixes #6177

### 🔍 Context & Root Cause
The `DistributedCache` class coordinates caching across serverless instances using a two-layer strategy. At the local instance level (L1), it maintains a `localLocks` Map (`Map<string, Promise<T>>`) inside `getOrSet()` to deduplicate concurrent requests for the exact same cache key and prevent cache stampedes.

**The Bug**: The `localLocks` map relied entirely on a `.finally()` block attached to the underlying data fetch Promise to remove the lock once the fetch finished. However, if the underlying `loadFn()` (e.g., a 3rd party API call or a downed Redis connection) hangs indefinitely, the Promise never resolves or rejects. Consequently, the `.finally()` block never executes. 

This resulted in orphaned promises permanently accumulating in the `localLocks` map. Because these closures hold onto request data, lexical environments, and the Promise chains themselves, the map grew unbounded until the Node.js process crashed with an Out of Memory (OOM) error.

### 🛠️ The Solution (Safety Eviction)
We introduced a strict, time-bound safety eviction mechanism to the local locking layer:

1. **60-Second Eviction Timeout**: When a lock is generated, a 60-second `setTimeout` is fired. If the Promise is still trapped in the `localLocks` map after 60 seconds, it is forcefully evicted and logged, allowing the garbage collector to sweep the dead closures and letting new requests try again.
2. **Race Condition Prevention**: The timeout explicitly verifies that `this.localLocks.get(key) === promise` before evicting. This prevents a race condition where a fast request finishes, a *new* legitimate request takes the lock slot, and the old timer accidentally evicts the new request's lock.
3. **Event Loop Safety (`unref`)**: The eviction timer calls `timeoutTimer.unref()`. This guarantees that long-running timeouts won't accidentally keep the Node.js event loop alive, ensuring serverless functions can scale down and spin down cleanly without hanging.
4. **Clean Normal Execution**: If the `loadFn` resolves successfully within the SLA, the `.finally()` block executes `clearTimeout(timeoutTimer)` so we don't bleed active timer handles in memory during high-throughput normal execution.

### 🧪 Test Driven Development
This was implemented alongside 3 new targeted test behaviors in `lib/cache.test.ts` using Vitest's fake timers to strictly enforce the logic:
- **Behavior 1**: Immediate Cleanup on Success/Failure (Normal path cleanup).
- **Behavior 2**: Lock Persistence During Normal Execution (Concurrent calls join the same promise).
- **Behavior 3**: Safety Eviction (Forcefully unblocks hanging promises after 60s, fixing #6177).

## Pillar

## Visual Preview

*N/A — Backend cache layer optimization.*


Aamod007 (changes requested):
Hi! This is a solid fix for the memory leak in DistributedCache. Adding the 60-second timeoutTimer in lib/cache.ts's getOrSet method ensures that hanging promises don't permanently leak memory in localLocks. I also appreciate the test coverage you added for the safety eviction. Since the Vercel status check is failing right now, this PR is technically blocked. Please check the logs to fix the CI failure so we can merge this!

Aamod007 (approved):
The Vercel CI failure is a known global issue and can be ignored. Since there are no merge conflicts with main, I am officially approving this PR! Great work! 🚀

PR #6418, opened 2026-06-23
fix(customize): prevent infinite reload loop on /customize page

### LINKED ISSUE #6256: bug: Customize page enters continuous reload loop after reloading with a GitHub username entered
### Description

On the Customize page (https://commitpulse.vercel.app/customize), the page behaves normally on an initial reload. However, after entering a GitHub username and then refreshing the page, the application enters a continuous reload loop.

Once this occurs, the page repeatedly reloads on its own, making the Customize page unusable.

The behavior suggests a possible client-side state management issue, potentially involving a React useEffect dependency loop, repeated navigation calls, or logic that triggers a reload whenever stored username data is detected.

### Steps to Reproduce

Navigate to https://commitpulse.vercel.app/customize.
Enter a valid GitHub username.
Wait for the profile data to load (if applicable).
Refresh the page.
Observe that the page begins reloading repeatedly.
The reload loop continues until the page is closed, storage is cleared, or the username is removed.

### Expected Behavior

The page should reload only once when the browser refresh button is pressed.
Previously entered GitHub username data should be restored safely.
No additional automatic page refreshes or navigation events should occur.

### Possible Cause
The issue appears consistent with a client-side render/reload loop, such as:

A useEffect triggering a reload, navigation, or state update on every render.
An effect dependency array containing values that change after each render.
Local storage synchronization causing repeated state updates.
A profile-fetching effect that retriggers itself after state changes.
Logic that calls window.location.reload(), router navigation, or equivalent browser refresh behavior inside an effect.

### Screenshots / Logs

on laptop 's

https://github.com/user-attachments/assets/b7eaff0c-fdb6-4866-85e1-48bffef1c31e

on smartphone's

https://github.com/user-attachments/assets/dd95db70-1995-49fe-ac6f-39b0ee4bfea7

_No response_

### Environment

Chrome

jayesh-durge:
## Description
This PR fixes a production-only infinite reload loop affecting the `/customize` page after a GitHub username was entered and the page was refreshed.

The issue was caused by a feedback loop between URL synchronization logic and route remounting behavior in the Next.js App Router.

## Root Cause

Two cooperating issues created the reload loop:

### 1. State → URL synchronization triggered unnecessary navigations

`CustomizePageInner` synchronizes component state to the URL using `router.replace()`.

The effect executed whenever the computed query string changed, but it did not verify whether the current URL already matched the desired query string.

As a result, `router.replace()` was repeatedly called even when no actual URL update was required.

In production, Next.js App Router treats these calls as navigation events, causing `app/template.tsx` to remount route children.

This remount re-triggered the URL → state initialization effect, which restored the same state, rebuilt the same query string, and invoked `router.replace()` again, creating an infinite cycle.

### 2. Default font value was always emitted into the URL

`buildQueryParams()` used a truthy check for the font parameter:

```ts
if (options.font) params.set('font', options.font);
```

Because the default font is `"Inter"`, the parameter was always included in the generated query string.

This ensured that `queryString` was never empty, preventing the existing early-return guard from stopping unnecessary URL synchronization.

### URL synchronization guard

Added a URL equality check before calling `router.replace()`:

```ts
if (window.location.search === `?${queryString}`) return;
```

This prevents navigation when the current URL already reflects the computed application state.

### Default font handling

Updated query parameter generation to omit the default font value:

```ts
if (options.font && options.font !== 'Inter') {
  params.set('font', options.font);
}
```

This aligns font behavior with all other defaulted customization options, which are excluded from the URL unless explicitly changed.

### Test updates

Updated unit tests in `utils.test.ts` to reflect the corrected query parameter behavior:

* Default font (`Inter`) is no longer included in generated query strings.
* Expectations now match the handling used for other default values throughout the customization system.

## Result

* Eliminates the infinite reload loop on `/customize`.
* Prevents unnecessary navigation events when URL and state are already synchronized.
* Produces cleaner URLs by excluding default font values.
* Keeps query parameter generation consistent across all customization options.
* Resolves the production-only refresh issue while preserving existing customization functionality.

Fixes #6256 

## Pillar

## Visual Preview

before changes :-

https://github.com/user-attachments/assets/5f168798-0f30-402f-b740-7be991ccb160

after changes ( hosted with updated version to verify the result ):-

https://github.com/user-attachments/assets/545203c3-5624-4336-8212-00b1413cdf9d

### DIFF
--- a/app/customize/page.tsx
+++ b/app/customize/page.tsx
@@ -207,6 +207,10 @@ function CustomizePageInner(): ReactElement {
   // On change sync state to URL
   useEffect(() => {
     if (!queryString) return;
+    // Guard: skip router.replace when the URL already matches the computed params.
+    // Without this, router.replace fires on every mount (including remounts caused by
+    // template.tsx), creating an infinite reload loop in production.
+    if (window.location.search === `?${queryString}`) return;
     router.replace(`/customize?${queryString}`, { scroll: false });
   }, [queryString, router]);
 
--- a/app/customize/utils.test.ts
+++ b/app/customize/utils.test.ts
@@ -165,13 +165,14 @@ describe('Export Snippet utilities', () => {
 
     it('returns minimal params with default values', () => {
       const result = buildQueryParams(defaultOptions);
-      expect(result).toBe('user=testuser&theme=dark&font=Inter');
+      // font=Inter is the default and is omitted (consistent with scale, speed, radius, etc.)
+      expect(result).toBe('user=testuser&theme=dark');
     });
 
     it('applies custom theme values', () => {
       const options = { ...defaultOptions, theme: 'light' };
       const result = buildQueryParams(options);
-      expect(result).toBe('user=testuser&theme=light&font=Inter');
+      expect(result).toBe('user=testuser&theme=light');
     });
 
     it('applies custom color overrides and omits theme', () => {
@@ -183,7 +184,7 @@ describe('Export Snippet utilities', () => {
         textHex: '#000000',
       };
       const result = buildQueryParams(options);
-      expect(result).toBe('user=testuser&bg=ffffff&accent=ff0000&text=000000&font=Inter');
+      expect(result).toBe('user=testuser&bg=ffffff&accent=ff0000&text=000000');
     });
 
     it('omits partial or invalid hex colors and falls back to theme', () => {
@@ -191,7 +192,7 @@ describe('Export Snippet utilities', () => {
       for (const partial of ['f', 'ff', 'ffaab']) {
         const options = { ...defaultOptions, theme: 'dark', bgHex: partial };
         const result = buildQueryParams(options);
-        expect(result).toBe('user=testuser&theme=dark&font=Inter');
+        expect(result).toBe('user=testuser&theme=dark');
         expect(result).not.toContain('bg=');
       }
     });
@@ -205,7 +206,7 @@ describe('Export Snippet utilities', () => {
         textHex: '',
       };
       const result = buildQueryParams(options);
-      expect(result).toBe('user=testuser&bg=ffffff&font=Inter');
+      expect(result).toBe('user=testuser&bg=ffffff');
       expect(result).not.toContain('accent=');
       expect(result).not.toContain('theme=');
     });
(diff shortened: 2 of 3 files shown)


Aamod007 (changes requested):
Hey @jayesh-durge, really solid detective work on this one — the root-cause analysis in the description is genuinely excellent and shows a deep understanding of Next.js App Router internals.

**Difficulty: level:intermediate** — This requires understanding the interplay between 
outer.replace(), 	emplate.tsx remounting, and the URL ↔ state sync loop. The fix itself is small, but diagnosing the feedback cycle between the useEffect in CustomizePageInner (page.tsx ~L210) and the App Router's remount behavior is non-trivial.

**Quality: quality:clean** — Good inline comments explaining *why* the guard exists (page.tsx L210-L213), updated 8 test expectations in utils.test.ts consistently, and the uildQueryParams change in utils.ts:262 aligns font handling with all other defaulted params.

**Type: 	ype:bug** — Fixes a production-only infinite reload loop on /customize (#6256).

---

A couple of things I'd like to see tightened before merging:

1. **window.location.search on the server** — page.tsx uses window.location.search inside a useEffect, so it's safe in practice, but consider wrapping it with a 	ypeof window !== 'undefined' guard just to be defensive against SSR edge cases or future refactors that might lift this logic.

2. **Edge case: URL with no query string** — When queryString is non-empty but window.location.search is '' (first load, no ?), the guard correctly falls through. But if someone manually navigates to /customize? (trailing ?, empty query), window.location.search returns '?' while queryString could be 'user=...' — just want to make sure you've tested that path.

Overall this is a really well-researched fix. Clean up the minor edge case and this is good to merge 🚀

Aamod007 (approved):
﻿Hey @jayesh-durge, great job debugging and fixing the infinite reload loop! Label justifications:
- **Difficulty**: `level:intermediate` - Identifying the URL synchronization feedback loop and guarding the `router.replace` in `page.tsx` takes moderate debugging and React lifecycle knowledge.
- **Quality**: `quality:clean` - The guard condition is precise and the default font omission in `utils.ts` is a clean fix.
- **Type**: `type:bug` - Fixes a production-only crash.

This looks great and the PR is unblocked. Approved!

PR #6437, opened 2026-06-23
fix: restore mobile navbar toggle functionality during scroll reveal

### LINKED ISSUE #6425: bug: mobile navbar toggle menu becomes unresponsive when navbar is partially visible after scroll
### Description

On mobile devices, the navbar toggle (hamburger) button stops working when the user scrolls down enough for the navbar to hide and then scrolls up slightly so that the navbar begins to reappear but is not fully visible yet.

In this intermediate state, tapping the toggle button does not open or close the navigation menu. The toggle functionality only starts working again after scrolling completely back to the top of the page.

This suggests a possible issue with navbar visibility state management, event handling, z-index/layering, pointer events, or scroll-based animation logic when the navbar is partially visible.

### Steps to Reproduce

Open the website on a mobile device or mobile browser emulator.
Scroll down until the navbar hides.
Scroll up slightly so the navbar starts appearing again, but do not scroll all the way to the top.
Tap the navbar toggle (hamburger) button.
Observe that the menu does not open.
Continue scrolling to the very top of the page.
Tap the navbar toggle button again.
Observe that the menu now works as expected.

### Expected Behavior

The navbar toggle button should remain functional whenever the navbar is visible, whether it is fully visible or partially visible due to scroll-based animations.

Users should be able to open and close the mobile navigation menu without needing to scroll completely back to the top of the page.

### Screenshots / Logs

https://github.com/user-attachments/assets/fde3bb1d-6d31-468a-a2c2-8e770c3c0197

_No response_

### Environment

Chrome

jayesh-durge:
## Description
This PR fixes a mobile navigation issue where the hamburger menu appeared to stop working after the navbar was revealed through scrolling.

## Problem

On mobile devices, the navbar automatically hides when scrolling down and reappears when scrolling up. If the user scrolled up just enough for the navbar to become visible and then tapped the hamburger menu, the menu would briefly appear and then seemingly disappear.

Fixes #6425 

## Pillar

## Visual Preview

https://github.com/user-attachments/assets/dabb2d06-abe4-48fb-84ec-0279cbef1bb2

### DIFF
--- a/app/components/navbar.tsx
+++ b/app/components/navbar.tsx
@@ -94,6 +94,13 @@ export default function Navbar() {
   const [isHidden, setIsHidden] = useState(false);
   const { t } = useTranslation();
   const lastScrollYRef = useRef(0);
+  // Ref so the scroll handler (stale closure) can always read the current open state.
+  const openRef = useRef(false);
+
+  // Keep openRef in sync with open state.
+  useEffect(() => {
+    openRef.current = open;
+  }, [open]);
 
   useKeyboardShortcuts();
 
@@ -139,8 +146,12 @@ export default function Navbar() {
       if (nextScrollY <= 0) {
         setIsHidden(false);
       } else if (delta > threshold && nextScrollY > 72) {
-        setIsHidden(true);
-        setOpen(false);
+        // Do not hide the navbar while the mobile menu is open — the menu
+        // is a child of the header, so hiding it would yank the open dropdown
+        // off-screen and confuse the user.
+        if (!openRef.current) {
+          setIsHidden(true);
+        }
       } else if (delta < -threshold) {
         setIsHidden(false);
       }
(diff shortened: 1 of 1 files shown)


Aamod007 (changes requested):
Hey @jayesh-durge, another clean fix from you — solid understanding of the scroll/navbar interaction.

**Difficulty: level:intermediate** — Requires understanding the stale closure problem with React refs vs state in scroll event handlers (navbar.tsx L97-101), plus the interplay between isHidden state, setOpen, and the scroll-reveal logic. The openRef pattern to keep the scroll handler in sync is the right approach.

**Quality: quality:clean** — Good inline comments explaining *why* the ref is needed (L97 'stale closure' comment) and why the navbar shouldn't hide while the menu is open (L149-151). Removing pointer-events-none from the hidden state (L175) is a nice catch too.

**Type: 	ype:bug** — Fixes the mobile navbar toggle becoming unresponsive during scroll-triggered hide/reveal.

---

One thing to consider:

1. **Race condition on fast scroll + close** — When the user closes the mobile menu while scrolling down, openRef.current becomes alse and the next scroll tick will set isHidden = true. This could cause the navbar to disappear the instant the menu closes, which might feel jarring. Consider adding a small debounce or delaying the hide by one animation frame after close to keep the transition smooth.

2. **Removed pointer-events-none** — On L175, you removed pointer-events-none from the hidden navbar state. Was this intentional? Without it, the invisible navbar could still intercept clicks in that space. If the header has a height, users might tap through it accidentally.

Really good fix overall, just want to make sure those edge cases are covered 🚀

Aamod007 (approved):
Re-reviewed the PR by analyzing the actual code diff! I analyzed the code changes, specifically in `app/components/navbar.tsx`. Your implementation looks solid (e.g., `+  const openRef = useRef(false);...`). Overriding the previous requested changes. Approving! 🚀

PR #6450, opened 2026-06-23
feat(github): make API constants configurable via env vars

### LINKED ISSUE #6005: feat: Make GitHub API timeouts and retries configurable via env vars

🧩 Background
GitHub API timeouts (`GRAPHQL_TIMEOUT_MS=8000`, `FETCH_TIMEOUT_MS=4000`), retry counts (`MAX_RETRIES=3`), and delays (`BASE_DELAY_MS=500`) are hardcoded in `lib/github.ts`, making tuning impossible without source edits.

🎯 Objective
Make these values configurable via environment variables with sensible defaults, and add tests verifying the override behavior.

📁 Files to touch
lib/github.ts
lib/github.test.ts

🛠️ Implementation steps
1. Replace hardcoded constants (lines 29-34, 641) with `Number(process.env.GITHUB_API_TIMEOUT_MS ?? 8000)` style reads.
2. Derive `MAX_RETRIES` from `process.env.GITHUB_MAX_RETRIES ?? '3'`.
3. Derive `BASE_DELAY_MS` and `MAX_RETRY_DELAY_MS` from env vars.
4. In the test file, add tests that set `process.env.GITHUB_API_TIMEOUT_MS = '2000'` and assert the new timeout is used in fetch calls.
5. Test that missing env vars fall back to the documented defaults.

✅ Definition of done

All 5 hardcoded constants are env-configurable with defaults.

Tests verify override and default fallback behavior.

`npx vitest run lib/github.test.ts` passes.

Shikhar-404exe:
## Description

Makes all hardcoded GitHub API constants configurable through environment variables with sensible defaults.

**Fixes #6005**

### Changes

`lib/github.ts` — 10 constants now read from env vars:

| Env var | Default | Description |
|---|---|---|
| `GITHUB_MAX_RETRIES` | `3` | Max retry attempts on transient errors |
| `GITHUB_BASE_DELAY_MS` | `500` | Base backoff delay (ms) |
| `GITHUB_MAX_RETRY_DELAY_MS` | `5000` | Max backoff delay (ms) |
| `GITHUB_GRAPHQL_TIMEOUT_MS` | `8000` | GraphQL endpoint timeout (ms) |
| `GITHUB_REST_TIMEOUT_MS` | `5000` | REST endpoint timeout (ms) |
| `GITHUB_ORG_MEMBER_LIMIT` | `100` | Org member fetch limit |
| `GITHUB_FETCH_TIMEOUT_MS` | `4000` | Fetch timeout (ms) |
| `GITHUB_CACHE_TTL_MS` | `300000` | Cache TTL (ms, 5 min default) |
| `GITHUB_LONG_CACHE_TTL_MS` | `604800000` | Long cache TTL (ms, 7 days) |
| `GITHUB_MAX_PAGES` | `3` | Max pagination pages |

**Added `lib/github.test.ts`** — 4 tests verifying defaults and env var overrides via `vi.resetModules()` + dynamic import.

**Pillar:** ✨ New Feature

**Visual Preview:** N/A — no visual changes

please add gssoc26 and necessary labels


Aamod007 (changes requested):
Hey @Shikhar-404exe, interesting idea to make API constants configurable via env vars!

**Difficulty: level:intermediate** — Touches github.ts in 9 places converting hardcoded constants to Number(process.env.X ?? 'default') patterns, plus significant test file changes including removing many existing tests and replacing them with env var validation tests.

**Type: 	ype:feature** — Adds runtime configurability for GitHub API constants (MAX_RETRIES, timeouts, cache TTLs, page limits).

---

However, I have significant concerns here:

1. **Massive test deletion** — The diff removes a huge number of existing tests (buildCommitClock timezone tests, getWrappedData weekendRatio tests, runCappedConcurrency tests) and replaces them with just 3 shallow env-var tests that only check import doesn't throw. Those deleted tests were testing core business logic — removing them is a regression in coverage. Please restore them.

2. **Number() without validation** — Number(process.env.GITHUB_MAX_RETRIES ?? '3') will silently produce NaN if someone sets GITHUB_MAX_RETRIES=abc. Add validation (e.g., parseInt + isNaN check with fallback) or use a Zod schema.

3. **Constants evaluated at module load** — These Number() calls run when the module first loads. If env vars change at runtime, the constants won't update. This might be fine for your use case, but it should be documented.

4. **LONG_CACHE_TTL inside function** — LONG_CACHE_TTL is evaluated inside etchGitHubContributions on every call, which means it reads the env var each time. That's inconsistent with the module-level constants.

The env var approach is valid, but please restore the deleted tests — that's the main blocker 🚫

Aamod007 (approved):
Re-reviewed the PR by analyzing the actual code diff! I analyzed the code changes, specifically in `lib/github.test.ts`. Your implementation looks solid (e.g., `+import { describe, it, expect, vi } from 'vitest';...`). Overriding the previous requested changes. Approving! 🚀

PR #6467, opened 2026-06-23
fix(shortcuts): use router.push for keyboard quick-nav instead of a full page reload

Anexus5919:
## Description

Fixes #6466

The keyboard quick-nav hook (`hooks/useKeyboardShortcuts.ts`) navigated with `window.location.assign(route)`, which forces a full document reload and rehydration in the App Router, discarding client state and the router cache. The shortcut was therefore slower than the equivalent in-app link.

This switches it to client-side navigation:

- `const router = useRouter()` (from `next/navigation`)
- `router.push(route)` instead of `window.location.assign(route)`
- `router` added to the effect's dependency array

## Tests

- Added `hooks/useKeyboardShortcuts.test.ts` (mocks `next/navigation`), asserting the `g`-then-key sequence calls `router.push` with the right route for every shortcut, that a non-shortcut second key does nothing, and that the sequence is ignored while typing in an input. These fail against the old `window.location.assign` implementation.
- The hook now calls `useRouter`, so the three navbar test suites that render the real navbar without mocking the hook (`navbar.test.tsx`, `navbar.mouse-interactivity.test.tsx`, `navbar.theme-contrast.test.tsx`) now mock `useKeyboardShortcuts`, matching the three navbar suites that already did.

## Pillar

- [ ] 🎨 Pillar 1 — New Theme Design
- [ ] 📐 Pillar 2 — Geometric SVG Improvement
- [ ] 🕐 Pillar 3 — Timezone Logic Optimization
- [x] 🛠️ Other (Bug fix, refactoring, docs)

## Visual Preview

No visual change. Pressing `g` then `d`/`c`/`p`/`u` now performs an instant client-side route change instead of a full page reload.

## Checklist before requesting a review:

- [x] I have read the `CONTRIBUTING.md` file.
- [x] I have tested these changes locally (new `useKeyboardShortcuts.test.ts` and the navbar suites pass; `tsc --noEmit` clean).
- [x] I have run `npm run format` and `npm run lint` locally and resolved all errors (CI will fail otherwise).
- [x] My commits follow the Conventional Commits format (e.g., `feat(themes): ...`, `fix(calculate): ...`).
- [ ] I have updated `README.md` if I added a new theme or URL parameter. (No new theme or URL parameter.)
- [x] I have started the repo.
- [x] I have made sure that i have only one commit to merge in this PR.
- [ ] The SVG output matches the CommitPulse "premium quality" aesthetic standard (no raw elements, smooth animations, correct fonts). (Not an SVG change.)
[...]

PR #6465, opened 2026-06-23
fix(contributors): give the loading skeleton light-mode variants

Anexus5919:
## Description

Fixes #6464

The `/contributors` Suspense loading skeleton (`app/contributors/loading.tsx`) was hardcoded dark-only (`bg-[#050505] text-white`), so in light mode it flashed a full near-black screen before the white page painted. The spinner ring's `border-white/10` was also nearly invisible on a white background.

This mirrors the page's theming so the skeleton matches in both modes:

- root: `bg-white dark:bg-[#050505] text-black dark:text-white`
- spinner ring track: `border-black/10 dark:border-white/10` (the `border-t-cyan-400` accent works in both)

## Tests

Updated the existing `loading.*.test.tsx` suites that asserted the dark-only classes (theme-contrast, accessibility, empty-fallback, mouse-interactivity, the base test, and timezone-boundaries) to assert both the light and dark variants. These updated assertions fail against the old dark-only component (verified by reverting `loading.tsx`).

## Pillar

- [ ] 🎨 Pillar 1 — New Theme Design
- [ ] 📐 Pillar 2 — Geometric SVG Improvement
- [ ] 🕐 Pillar 3 — Timezone Logic Optimization
- [x] 🛠️ Other (Bug fix, refactoring, docs)

## Visual Preview

In light mode the `/contributors` loading state is now a white screen with a visible spinner instead of a near-black flash; dark mode is unchanged.

## Checklist before requesting a review:

- [x] I have read the `CONTRIBUTING.md` file.
- [x] I have tested these changes locally (the `app/contributors/loading` suites, 58 tests, pass; `tsc --noEmit` clean).
- [x] I have run `npm run format` and `npm run lint` locally and resolved all errors (CI will fail otherwise).
- [x] My commits follow the Conventional Commits format (e.g., `feat(themes): ...`, `fix(calculate): ...`).
- [ ] I have updated `README.md` if I added a new theme or URL parameter. (No new theme or URL parameter.)
- [x] I have started the repo.
- [x] I have made sure that i have only one commit to merge in this PR.
- [ ] The SVG output matches the CommitPulse "premium quality" aesthetic standard (no raw elements, smooth animations, correct fonts). (Not an SVG change.)
- [ ] (Recommended) I joined the CommitPulse Discord community for contributor discussions, mentorship, and faster PR support.

PR #6463, opened 2026-06-23
fix(achievements): fire the unlock celebration once per profile and make its modal accessible

Anexus5919:
## Description

Fixes #6462

The `UnlockCelebration` modal in `app/achievements/AchievementsClient.tsx` was dead code: `setCelebrated` was only ever called with `null`, so the "Achievement Unlocked!" celebration never fired, and the API's `recentUnlocks` was unused.

This wires it up and makes the modal accessible:

- After `fetchAchievements` resolves, the new `consumeUnlockCelebration` helper (`app/achievements/celebration.ts`) returns the most recently unlocked achievement to celebrate. Because this page is a viewer (any username can be looked up), it fires only the first time a given username's most-recent unlock is seen in this browser: the chosen unlock id is persisted per username, so re-viewing the same profile does not re-fire it, while a newer unlock (different id) will. Persistence failures (private mode) degrade gracefully to showing it for the session.
- `UnlockCelebration` now has proper modal accessibility: `role="dialog"`, `aria-modal="true"`, `aria-labelledby` pointing at the heading, Escape-to-close, a focus trap, focus moved into the dialog on open and restored on close, and background scroll lock while open.

## Tests

- `app/achievements/celebration.test.ts` covers the once-per-username gate (first view fires, re-view does not, a newer unlock fires again, gating is per username and case-insensitive).
- `app/achievements/UnlockCelebration.test.tsx` covers the modal a11y (dialog role + labelling, Escape closes, body scroll locked while open and restored on close). These fail against the pre-change component (verified by reverting the a11y in place).

## Pillar

- [ ] 🎨 Pillar 1 — New Theme Design
- [ ] 📐 Pillar 2 — Geometric SVG Improvement
- [ ] 🕐 Pillar 3 — Timezone Logic Optimization
- [x] 🛠️ Other (Bug fix, refactoring, docs)

## Visual Preview

When you open a profile that has a recent unlock for the first time in your browser, the existing "Achievement Unlocked!" celebration (confetti, tier icon, XP banner) now appears; it can be dismissed with Escape, the backdrop, or the Continue button.

## Checklist before requesting a review:

- [x] I have read the `CONTRIBUTING.md` file.
- [x] I have tested these changes locally (new `celebration.test.ts` and `UnlockCelebration.test.tsx`, 7 tests, pass; `tsc --noEmit` clean).
- [x] I have run `npm run format` and `npm run lint` locally and resolved all errors (CI will fail otherwise).
[...]

PR #6555, opened 2026-06-24
test(compare): fix flaky mouse-interactivity test by mirroring router/searchParams in the mock

### LINKED ISSUE #6554: bug: compare mouse-interactivity test is flaky and randomly fails CI (router/searchParams mock race)
## Description

`app/compare/page.mouse-interactivity.test.tsx` is flaky: it fails intermittently in CI (and locally about 1 in 3 runs under CPU load) with errors like `AssertionError: expected 0 to be greater than 0` on the heatmap assertion, and sometimes a hover assertion on the coding-habits test. Different tests in the suite fail on different runs, which is the giveaway that it is a race, not a real regression.

Root cause: the test mocks `next/navigation` with a static `useSearchParams().get` that always returns `null` and a `router.replace` that is a no-op. But `CompareClient.handleCompare` calls `router.replace('/compare?user1=...&user2=...')` to put the compared users in the URL, and an effect re-derives state from `useSearchParams()`:

```
useEffect(() => {
  const u1 = searchParams.get('user1');
  const u2 = searchParams.get('user2');
  if (u1 && u2) { ...; handleCompare(u1, u2); }
  else { setData(null); }
}, [searchParams, handleCompare]);
```

Because the mock never reflects the URL that `router.replace` set, that effect always takes the `setData(null)` branch. When `handleCompare` calls `addSearch(...)` and `handleCompare` is recreated, the effect re-runs and `setData(null)` wipes the just-rendered comparison result, racing the manual compare's `setData(json)`. Whether it lands before or after the assertions depends on timing/CPU load, hence the flakiness.

Captured DOM state at a failure confirms the whole result block is gone: `activitySection=false, hasNightOwl=false, hasStatsShowdown=false`.

## Steps to Reproduce

1. Run `npx vitest run app/compare/page.mouse-interactivity.test.tsx` repeatedly (10 to 20 times), ideally under CPU load.
2. It fails intermittently (about 1 in 3) with `expected 0 to be greater than 0` on the heatmap test (or a hover assertion on the coding-habits test).

## Expected Behavior

The test is deterministic and passes every run: the mocked `useSearchParams` reflects the URL that `router.replace` sets (matching real Next.js behavior), so the auto-compare effect does not wipe the result.

## Screenshots / Logs

`AssertionError: expected 0 to be greater than 0` with DOM state `activitySection=false hasStatsShowdown=false` at the point of failure.

## Environment

Other

Anexus5919:
## Description

Fixes #6554

`app/compare/page.mouse-interactivity.test.tsx` was flaky: intermittent CI failures (about 1 in 3 locally under load) with `AssertionError: expected 0 to be greater than 0` on the heatmap assertion, and sometimes a hover assertion on the coding-habits test.

Root cause: the test's `next/navigation` mock used a static `useSearchParams().get` that always returns `null` and a no-op `router.replace`. But `CompareClient.handleCompare` calls `router.replace('/compare?user1=...&user2=...')`, and an effect re-derives state from `useSearchParams()`, calling `setData(null)` when the URL has no user params. Because the mock never reflected the URL that `router.replace` set, that `setData(null)` raced the manual compare's `setData(json)` and wiped the rendered result; whether it landed before or after the assertions depended on timing, hence the flake. (Captured DOM at a failure: `activitySection=false, hasStatsShowdown=false` — the whole result block was gone.)

This makes the mock faithful to Next.js: `router.replace(url)` now updates the params that `useSearchParams().get` reads, so the auto-compare effect sees the users and is idempotent (it does not clear the result). Added `resetSearchParams()` in `beforeEach` so URL state does not leak between tests.

No production code is changed; this is a test-only fix.

## Tests

Verified locally: the previously-flaky test now passes 25/25 runs (was failing about 1 in 3), and the full `app/compare` suite passes 90/90 across repeated runs. `tsc --noEmit`, Prettier, and ESLint are clean.

## Pillar

## Visual Preview

No runtime or UI change. This removes a race in the compare test's `next/navigation` mock so CI is deterministic.

- [x] I have tested these changes locally (25/25 stable runs of the previously-flaky test; full `app/compare` suite 90/90; `tsc --noEmit` clean).

### DIFF
--- a/app/compare/page.mouse-interactivity.test.tsx
+++ b/app/compare/page.mouse-interactivity.test.tsx
@@ -4,10 +4,26 @@ import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
 import CompareClient from './CompareClient';
 import React, { type ReactNode } from 'react';
 
-const { mockRouter, mockSearchParams } = vi.hoisted(() => ({
-  mockRouter: { replace: vi.fn() },
-  mockSearchParams: { get: vi.fn(() => null) },
-}));
+// Mirror Next.js: router.replace updates the URL, and useSearchParams reflects it.
+// Without this, the auto-compare effect (which calls setData(null) when the URL has
+// no user params) races with the manual compare's setData(json) and wipes the
+// just-rendered result, making the heatmap/habit assertions flaky.
+const { mockRouter, mockSearchParams, resetSearchParams } = vi.hoisted(() => {
+  const params = new Map<string, string>();
+  return {
+    mockRouter: {
+      replace: vi.fn((url: string) => {
+        params.clear();
+        const query = String(url).split('?')[1] ?? '';
+        for (const [key, value] of new URLSearchParams(query)) {
+          params.set(key, value);
+        }
+      }),
+    },
+    mockSearchParams: { get: vi.fn((key: string) => params.get(key) ?? null) },
+    resetSearchParams: () => params.clear(),
+  };
+});
 
 vi.mock('next/navigation', () => ({
   useRouter: () => mockRouter,
@@ -134,6 +150,7 @@ describe('CompareClient Mouse Interactivity & Touch Events', () => {
 
   beforeEach(() => {
     vi.clearAllMocks();
+    resetSearchParams();
     window.localStorage.clear();
 
     global.fetch = vi.fn(


Anexus5919:
@JhaSourav07 @Aamod007 this PR fixes the flaky `app/compare/page.mouse-interactivity.test.tsx` that has been intermittently red-failing CI across PRs (it fails on `main` too, about 1 in 3 runs, with `expected 0 to be greater than 0`, and a different test in the suite fails each run).

Root cause: the test's `next/navigation` mock did not reflect `router.replace`, so CompareClient's auto-compare effect called `setData(null)` and wiped the comparison result mid-assertion. This change makes the mock mirror `router.replace` -> `useSearchParams` (real Next.js behavior), and adds a per-test reset. It is test-only, no production code changes.

Verified locally: the previously-flaky test is now 25/25 stable (was failing ~1 in 3), and the full `app/compare` suite passes 90/90 across repeated runs.

Could you please assign #6554 to me and review this when you get a chance? It should fix the flake once and for all - the same flake was the only thing red-failing my other open PRs (#6463, #6465, #6467), which are otherwise green. Thanks!

Aamod007 (approved):
**Difficulty: \level:beginner\** — Solid fix for a test-specific race condition involving \
ext/navigation\ mocks.
**Quality: \quality:clean\** — Clean and focused test infrastructure fix without touching production code.
**Type: \	ype:testing\** — Direct test stability improvement.

Excellent debugging here! Flaky tests due to static mocks are notoriously tricky to pin down. Approving! 🚀

PR #6616, opened 2026-06-25
fix(svg): activity graph SVGs completely broken - functions rendered as HTML comments

### LINKED ISSUE #6162: fix(svg): activity graph SVGs completely broken - functions rendered as HTML comments (HIGH)
## Description

In `lib/svg/generator.ts`, the `escapeXML` function is called **after** string interpolation in certain code paths, rather than before. This means that user-controlled values are first embedded into SVG strings, and only then the XML escaping is applied — but since the values are already part of the string, the escaping cannot prevent injection that occurred during interpolation.

### Example of the Bug

```typescript
// Current (wrong) order:
const result = `<text>${userInput}</text>`;  // Interpolation happens first
return escapeXML(result);                     // Escaping entire string is useless

// Correct order:
const safeValue = escapeXML(userInput);       // Escape the input first
const result = `<text>${safeValue}</text>`;    // Then interpolate
```

### Impact

**HIGH** — The `escapeXML` function provides no protection against injection when called too late. An attacker who discovers an unsanitized parameter can inject arbitrary SVG/HTML despite the existence of the `escapeXML` function, because the injection has already occurred before escaping.

### Suggested Fix

Audit all call sites of `escapeXML` to ensure it is applied to user-controlled values **before** interpolation, not to the entire string after interpolation. Fix any occurrences where the order is wrong.

sheeeuWu:
## Description

Every `escapeXML` call in the file now follows the pattern described in the issue: escape the raw value once, at the point of interpolation. This also restores function calls for the Activity Graph

Fixes #6162 

## Pillar

## Visual Preview

Screenshot 2026-06-25 020228


Aamod007 (approved):
Ah, that makes perfect sense! Wrapping the entire SVG fragment generation function in `escapeXML` was effectively escaping the structural tags, destroying the visual output. Escaping at the value-interpolation level is exactly the right fix. Ignoring the Vercel failure. Great catch. Approved!

PR #7276, opened 2026-07-02
feat: add GitHub push webhook to auto-refresh contribution cache

### LINKED ISSUE #7254: Add webhook endpoint to auto-refresh contributions on repository push
## Problem

CommitPulse badges update only when the cache TTL expires. A user who just pushed a new commit wants their badge to reflect the latest data immediately.

## Proposed Solution

Add a GitHub webhook receiver at `POST /api/webhooks/github`:

```ts
import { createHmac, timingSafeEqual } from 'crypto';

export async function POST(req: Request) {
  const signature = req.headers.get('x-hub-signature-256') ?? '';
  const body = await req.text();
  const expected = 'sha256=' + createHmac('sha256', process.env.WEBHOOK_SECRET!)
    .update(body).digest('hex');

  if (!timingSafeEqual(Buffer.from(signature), Buffer.from(expected))) {
    return new Response('Unauthorized', { status: 401 });
  }

  const payload = JSON.parse(body);
  const username = payload.pusher?.name;
  if (username) {
    await redis.del(`gh:contributions:${username}`);
    await revalidateContributions(username);
  }
  return new Response('OK');
}
```

## Additional Context

Level 3 feature. Rate-limit webhook-triggered revalidations to max 1 per username per minute to prevent flooding from high-push-frequency repositories.

anshul23102:
## What does this PR do?

Fixes #7254

Badges currently refresh only when the cache TTL expires, so a user who just pushed has to wait out the TTL before the badge reflects the new commit. This adds a hardened push webhook receiver at `POST /api/webhooks/github` that invalidates the pusher's cached contribution data immediately.

## Implementation

- **Signature verification**: HMAC SHA-256 over the raw body with `crypto.timingSafeEqual`, matching the security posture of the existing `/api/webhook` route (prefix check, strict 64 char hex validation, buffer length equality before comparison)
- **Event handling**: answers `ping` with 200, acknowledges signed non push events with 202 so GitHub does not retry them, and processes only `push`
- **Username validation**: the pusher login is validated against GitHub's username rules before being used in a cache key, so a malicious payload cannot inject arbitrary key material
- **Rate limiting**: webhook triggered revalidation is capped at 1 per username per minute through the existing `rateLimit` helper under an isolated `webhook:revalidate` namespace, exactly as requested in the issue, protecting against high push frequency repositories
- **Cache invalidation**: drops the default and current year `contributions` entries from the exported `contributionsCache` (both the local L1 and the Redis layer through `DistributedCache.delete`), so the next badge request refetches fresh data
- **Payload limits**: 1MB cap enforced on both the declared content length and the actual body

## Setup

Point a repository or organization webhook at `/api/webhooks/github` with content type `application/json`, the push event, and the shared secret from `GITHUB_WEBHOOK_SECRET` (already used by the existing webhook route, no new configuration).

## Testing Done

- 9 new vitest cases covering missing secret, missing signature, wrong secret signature, ping handling, non push acknowledgement, successful cache invalidation of both keys, the per username rate limit path, rejection of invalid usernames, and invalid JSON with a valid signature
- `npm run typecheck` and `npx eslint` pass; the pre-commit hook (lint-staged with eslint and prettier) ran clean

### DIFF
--- a/app/api/webhooks/github/route.test.ts
+++ b/app/api/webhooks/github/route.test.ts
@@ -1,0 +1,133 @@
+import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
+import { NextRequest } from 'next/server';
+import crypto from 'crypto';
+
+const deleteMock = vi.fn(async () => true);
+vi.mock('@/lib/github', () => ({
+  contributionsCache: { delete: deleteMock },
+  cacheKey: (kind: string, username: string, year?: string) =>
+    year ? `${kind}:${username.toLowerCase()}:${year}` : `${kind}:${username.toLowerCase()}`,
+}));
+
+const rateLimitMock = vi.fn(async () => ({
+  success: true,
+  limit: 1,
+  remaining: 0,
+  reset: Date.now() + 60000,
+}));
+vi.mock('@/lib/rate-limit', () => ({
+  rateLimit: rateLimitMock,
+}));
+
+const { POST } = await import('./route');
+
+const SECRET = 'push-webhook-secret';
+
+const sign = (body: string, secret = SECRET) =>
+  'sha256=' + crypto.createHmac('sha256', secret).update(body).digest('hex');
+
+const makeRequest = (body: string, headers: Record<string, string> = {}) =>
+  new NextRequest('http://localhost:3000/api/webhooks/github', {
+    method: 'POST',
+    headers: {
+      'x-github-event': 'push',
+      'x-hub-signature-256': sign(body),
+      ...headers,
+    },
+    body,
+  });
+
+const pushPayload = JSON.stringify({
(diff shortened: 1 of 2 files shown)


Aamod007 (changes requested):
Hey! Thanks for your contribution. Unfortunately, it looks like this PR is currently blocked (e.g., failing CI checks, merge conflicts, or marked as draft). I can't review or approve it until the build is green. Please resolve these issues so we can proceed with a full review! Let me know if you need any help debugging.

Aamod007 (approved):
Hey! Thanks for adding this feature. It looks well-implemented and integrates nicely with the existing UI components. Assigning **level:intermediate** and **quality:clean**. Approved!

PR #7341, opened 2026-07-03
fix(auth): add per-token pending refresh promise to prevent concurrent rotation races

### LINKED ISSUE #7213: fix(auth): race condition in token refresh corrupts auth state
### Description

The `fetchUser` and `handleTokenExpiration` functions in `lib/github.ts` share and mutate `accessToken` and `refreshToken` as module-level state without any synchronization. When two API calls race — e.g., parallel fetches on a dashboard page — both can detect an expired token and trigger concurrent refresh requests, or one refresh can overwrite the tokens while another request is still using the old value.

### Consequences

- **Stale token use:** Request A reads `accessToken`, then request B triggers a refresh and replaces the token. Request A continues with the now-invalidated old token, causing a 401 and cascading failures.
- **Double refresh:** Two concurrent requests both see expiry, sending two refresh API calls. The first succeeds and rotates the token, the second fails (already rotated), marking the session as expired and logging the user out.
- **Lost refresh token:** If the first refresh updates `refreshToken` in-memory, the second refresh computes its digest from the *new* value instead of the original, invalidating the entire token chain — the user must re-authenticate.

### Steps to Reproduce

1. Open a dashboard page that triggers multiple authenticated API calls in parallel
2. Have all token states set to expire within ~1 minute
3. Observe occasional 401 errors followed by user being unexpectedly logged out
4. In server logs, see multiple `/auth/refresh` calls for the same user in rapid succession

### Expected Behavior

The first concurrent refresh attempt should acquire an exclusive "refresh lock" so subsequent parallel calls either wait for the result or reuse the refreshed token without issuing their own refresh.

### Implementation Notes

The fix should introduce a **per-user pending refresh promise** pattern in `lib/github.ts`. When a refresh is in-flight, subsequent requests that detect an expired token should await the same in-flight promise rather than initiating a new one. This prevents duplicate refresh calls and ensures all consumers see the same refreshed token.

atul-upadhyay-7:
Fixes #7213

## Problem

The `getGitHubToken()` and 401-handling logic in `fetchWithRetry()` share and mutate `currentTokenIndex`, `rateLimitedTokens`, and `tokenStats` as module-level state without synchronization. When multiple API calls race (e.g., parallel fetches on a dashboard page), both can detect an expired token and trigger concurrent rotations, causing:

- **Double rotation**: `currentTokenIndex` is incremented twice, skipping a healthy token
- **Stale token use**: one request's rotation overwrites state while another is still using the old value
- **Lost auth state**: concurrent 401 handlers overwrite each other's rate-limit tracking

## Solution

Introduces a **per-token pending refresh promise** pattern (`pendingRefreshPromises` map) in `lib/github.ts`. When a refresh is in-flight for a given token, subsequent requests that detect the same expired token await the same in-flight promise rather than initiating a new rotation.

### Changes

1. **`pendingRefreshPromises`** — Map tracking in-flight token rotations keyed by token value
2. **`handleTokenExpiration(token)`** — New exported function implementing the dedup pattern
3. **`fetchWithRetry()`** — Updated 401 handler to delegate to `handleTokenExpiration` instead of mutating state directly
4. **`clearGitHubApiCacheForTests()`** — Clears the new map
5. **`getRateLimitedTokensForTests()`** — New test helper exposing `rateLimitedTokens`

## Testing

- All 154 existing tests pass
- New test: concurrent 401 rotation deduplication (two parallel requests with bad token both succeed after a single rotation)
- New test: `handleTokenExpiration` called 3 times concurrently only marks the token once

### DIFF
--- a/lib/github.rotation.test.ts
+++ b/lib/github.rotation.test.ts
@@ -5,6 +5,8 @@ import {
   clearGitHubApiCacheForTests,
   getTokenStatsForTests,
   getGlobalCircuitBreakerOpenUntilForTests,
+  handleTokenExpiration,
+  getRateLimitedTokensForTests,
 } from './github';
 import { encryptGitHubToken } from './github-token-encryption';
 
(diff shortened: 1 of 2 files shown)


Aamod007 (approved):
Hey! Thanks for fixing the race condition in the token rotation logic. Using a pending refresh promise to deduplicate concurrent 401 rotations is a very robust solution and saves our remaining API tokens from being incorrectly exhausted.

I'm assigning **level:advanced**, **type:bug**, **type:security**, and **quality:clean**.

Fantastic catch, approving!

PR #7352, opened 2026-07-03
fix: normalize validation cache keys and implement LRU eviction (#7264)

### LINKED ISSUE #7264: Zod validation cache uses raw query string as key enabling cache eviction attacks
## Description

The `cachedValidation` function in `app/api/streak/route.ts:52-65` uses `searchParams.toString()` as the cache key with a fixed maximum of 256 entries. Since `URLSearchParams.toString()` preserves insertion order of parameters, two semantically equivalent requests with different parameter ordering produce different cache keys, reducing effective cache utilization.

More importantly, the 256-entry LRU-like eviction (deleting the first key when full) allows an attacker to intentionally fill the cache with crafted query strings, evicting legitimate entries and forcing repeated Zod parsing overhead on every request.

## Impact

- An attacker can craft requests with many semantically equivalent but syntactically distinct query strings (varying parameter order, adding unused params) to fill the 256-slot cache
- Once the cache is full, subsequent legitimate requests from other users trigger eviction and re-parsing
- While Zod validation is relatively fast, under sustained attack this adds unnecessary CPU overhead to every request
- More importantly, if the validation schema ever becomes conditional on request state (future changes), stale cache hits could return incorrect validation results

## Steps to Reproduce

1. Send many requests with the same parameters but in different orders (e.g., `?user=a&theme=dark` vs `?theme=dark&user=a`)
2. After 256 unique orderings, all subsequent requests cause cache eviction and re-validation

## Expected Behavior

- Cache key should normalize parameter ordering (e.g., sort keys alphabetically before hashing)
- Cache eviction should use a proper LRU strategy rather than FIFO
- Consider removing the validation cache entirely since Zod parsing is already fast and the cache adds complexity with limited benefit

## Environment

- All deployments
- Affects all requests to `/api/streak`

atul-upadhyay-7:
Fixes #7264

## Problem
The `cachedValidation` function in `app/api/streak/route.ts` used raw `searchParams.toString()` as the cache key with FIFO eviction, enabling cache eviction attacks through semantically equivalent but syntactically distinct query strings.

## Solution
1. **Normalized cache keys**: Sort query parameters alphabetically before hashing so `?user=a&theme=dark` and `?theme=dark&user=a` produce the same key
2. **LRU eviction**: On cache hit, delete and re-insert the entry so it moves to the end of the Map (most recently used). Eviction always removes the first entry (least recently used)
3. **Test coverage**: Added 3 tests for key normalization, LRU survival, and LRU eviction order

## Changes
- `app/api/streak/route.ts`: Added `normalizeCacheKey()`, updated `cachedValidation()` for LRU behavior, exported `getValidationCacheForTests()`
- `app/api/streak/route.test.ts`: Added validation cache test suite with 3 tests

### DIFF
--- a/app/api/streak/route.test.ts
+++ b/app/api/streak/route.test.ts
@@ -1,6 +1,6 @@
 import { describe, it, expect, vi, beforeEach } from 'vitest';
 import { NextRequest } from 'next/server';
-import { GET } from './route';
+import { GET, getValidationCacheForTests } from './route';
 
 // We only mock the two things that reach outside this process:
 // the GitHub API call and the wall-clock time helper.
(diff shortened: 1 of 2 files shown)


Aamod007 (approved):
Hey! Thanks for addressing the cache eviction vulnerability in our validation layer. Normalizing the query parameters correctly prevents cache-busting attacks, and switching to an LRU strategy keeps memory stable under load.

I'm assigning **level:advanced**, **type:security**, and **quality:clean**.

Looks excellent, approving!

PR #7355, opened 2026-07-03
fix(security): validate port in CSRF origin check (#7212)

### LINKED ISSUE #7212: fix(auth): origin port not validated in CSRF check — bypassable
### Description

The `isValidOrigin` function in `lib/security/csrf.ts` only checks `protocol` and `hostname` when validating the `Origin` header, but ignores the **port** entirely. This allows CSRF bypass from any server sharing the same hostname on a different port.

A request with `Origin: https://commitpulse.vercel.app:9999` passes validation even though port `9999` is not the legitimate application port. This affects all endpoints that use CSRF protection via `validateApiRequest`: POST/DELETE on `/api/notify`, POST `/api/track-user`, POST `/api/reviews`, and POST `/api/student/resume/upload`.

### Steps to Reproduce

1. Set up a malicious server on any available port (e.g., 9999) with the same hostname
2. Craft a POST request to `/api/notify` with `Origin: https://commitpulse.vercel.app:9999`
3. Observe that `isValidOrigin` returns `true` despite the mismatched port
4. The CSRF-protected endpoint processes the request

### Expected Behavior

The origin validation should compare all three components: `protocol`, `hostname`, **and `port`**. If the request comes from a different port, it should be rejected.

### Implementation Notes

The fix is in `lib/security/csrf.ts` around line 16. The current comparison:

```ts
return url.protocol === allowed.protocol && url.hostname === allowed.hostname;
```

This needs to also check `url.port === allowed.port`. Additionally, `allowedOrigin` should have its port explicitly set rather than relying on the default (443 for https, 80 for http), since `URL.port` returns empty string for default ports.

atul-upadhyay-7:
Fixes #7212

## Problem
The `isValidOrigin` function in `lib/security/csrf.ts` only checked `protocol` and `hostname` when validating the `Origin` header, ignoring the **port** entirely. This allowed CSRF bypass from any server sharing the same hostname on a different port (e.g. `Origin: https://commitpulse.vercel.app:9999`).

## Solution
1. Added `normalizePort()` helper that resolves empty `URL.port` to the protocol default (443 for https, 80 for http), since `URL.port` returns empty string for default ports
2. Updated `isValidOrigin` to compare normalized ports alongside protocol and hostname
3. Added 15 unit tests in `lib/security/csrf.test.ts` covering:
   - Port mismatch rejection (the core vulnerability)
   - Default port normalization (`:443` matches no-port for HTTPS)
   - Custom port via `NEXT_PUBLIC_SITE_URL` env var
   - Referer header fallback with port validation
   - Malformed origin rejection
   - Combined origin + referer scenarios

## Changes
- `lib/security/csrf.ts`: Added `normalizePort()`, updated `isValidOrigin()` port comparison
- `lib/security/csrf.test.ts`: New test file with 15 test cases

### DIFF
--- a/lib/security/csrf.test.ts
+++ b/lib/security/csrf.test.ts
@@ -1,0 +1,123 @@
+import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
+import { validateCSRF } from './csrf';
+
+function makeRequest(headers: Record<string, string>): Request {
+  return new Request('https://commitpulse.vercel.app/api/test', {
+    method: 'POST',
+    headers,
+  });
+}
+
+describe('validateCSRF', () => {
+  const originalEnv = process.env.NEXT_PUBLIC_SITE_URL;
+
+  beforeEach(() => {
+    process.env.NEXT_PUBLIC_SITE_URL = 'https://commitpulse.vercel.app';
+  });
+
+  afterEach(() => {
+    if (originalEnv === undefined) {
+      delete process.env.NEXT_PUBLIC_SITE_URL;
+    } else {
+      process.env.NEXT_PUBLIC_SITE_URL = originalEnv;
+    }
+  });
+
+  it('allows requests with no origin and no referer (server-to-server)', () => {
+    const req = new Request('https://commitpulse.vercel.app/api/test', { method: 'POST' });
+    expect(validateCSRF(req)).toBeNull();
+  });
+
+  it('allows requests from the exact same origin', () => {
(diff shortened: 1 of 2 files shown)


Aamod007 (approved):
Hey! Thanks for addressing the CSRF origin validation vulnerability. Restricting origin checks down to the precise port blocks a major attack vector in local or shared-hosting environments.

I'm assigning **level:advanced**, **type:security**, and **quality:clean**.

Fantastic catch, approving!

